PR #12, opened 2017-02-19
Fix getSignalHeaders

jukka:
The self.n_channels variable does not exist, use self.signals_in_file instead.


holgern:
Thank you for the pull request

PR #15, opened 2017-05-12
Essentially corrected a typo in readme

zaccharieramzi:
I also verified that installing `pywavelets`was not enough to make pyedflib working.


holgern:
Thank you.

PR #28, opened 2018-02-25
set_number_of_annotation_signals

### LINKED ISSUE #27: set_number_of_annotation_signals method needed
Currently limited to only 1 annotation 'channel'.

It wasn't entirely clear, but the [edflib header](https://github.com/Teuniz/EDFlib/blob/master/edflib.h#L643-L652) suggests that there can only be one annotation per data record. If we have many annotations then we need more than that. Please correct me if I'm wrong.

cboulay:
Fixes #27

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -113,7 +113,7 @@ class EdfWriter(object):
         self.recording_start_time = datetime.now()
         self.birthdate = date(1900, 1, 1)
         self.duration = 1
-        self.number_of_annotations = 1
+        self.number_of_annotations = 1 if file_type in [FILETYPE_EDFPLUS, FILETYPE_BDFPLUS] else 0
         self.n_channels = n_channels
         self.channels = []
         self.sample_buffer = []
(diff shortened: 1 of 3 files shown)


cboulay:
The builds are getting the following errors `OSError: the file is not EDF(+) or BDF(+) compliant (Filesize)`. However, with this change, I'm able to save files with more than 1 annotation per second and I can load them using either this package or with mne. Do you have any insight into the error?

pyedflib/edfwriter.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, file_name, n_channels,
         self.recording_start_time = datetime.now()
         self.birthdate = date(1900, 1, 1)
         self.duration = 1
+        self.number_of_annotations = 1
```
holgern: annotations are only 1 for BDFplus and EDFplus.
For BDF and EDF, annotations ore 0.

holgern: Please add
if self.file_type == FILETYPE_EDFPLUS or self.file_type == FILETYPE_BDFPLUS:
self.number_of_annotations = 1
else:
self.number_of_annotations = 0

PR #50, opened 2019-01-17
set genders correctly

skjerns:
I've figured that pyEDFlib is a gender-bender.

According to definitions of edflib (and the function header in `writeedf`), 0=Female, 1=Male.

I've fixed it, now giving 'Female' or 'F' should also write a 0


skjerns:
I think the build failing has nothing to do with my code, could that be?

skjerns:
@holgern can you take a look?

holgern:
Fixed in Commit https://github.com/holgern/pyedflib/commit/ffcc82dbe9f543ed8284c2fe7f27810c1405200a

PR #55, opened 2019-05-01
Explicitly call close when back to back writers are created since del…

dthkao:
… indirectly and lazily calls __del__ which can result in a IO race condition.


dthkao:
friendly ping?

PR #56, opened 2019-05-01
Fix issues with test_EdfReader_Close_file and have it actually exerci…

dthkao:
…se/verify expected behavior.

I'm a little bit confused by the intent of commit 330ca683fd6ffaf284dd5ad385d5a32cba4c7c10, but I did notice that assert_raises is invalid syntax and also seems to be a no-op in the except clauses. Hopefully this is a little better?


dthkao:
friendly ping?

PR #57, opened 2019-05-02
Switch digital dtype to int32.

### LINKED ISSUE #54: PR #44 seems to be causing test failures
Was PR #44 tested in Python 3? When I run test_edfwriter.py in py3 I see:

`ValueError: Buffer dtype mismatch, expected 'int32_t' but got 'long'
`
This seems to stem from `blockwrite_digital_samples` in _pyedflib.pyx explicitly expecting int32, but np.int (equivalently the builtin int) in python has a version-dependent underlying dtype.

Does using np.int32 make more sense here?

dthkao:
Fixes #54

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -668,19 +668,19 @@ class EdfWriter(object):
             ind.append(0)
 
         sampleLength = 0
-        sampleRates = np.zeros(len(data_list), dtype=np.int)
+        sampleRates = np.zeros(len(data_list), dtype=np.int32)
         for i in np.arange(len(data_list)):
             sampleRates[i] = self.channels[i]['sample_rate']
             if (np.size(data_list[i]) < ind[i] + self.channels[i]['sample_rate']):
                 notAtEnd = False
             sampleLength += self.channels[i]['sample_rate']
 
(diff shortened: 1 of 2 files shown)


dthkao:
friendly ping?

snbarr:
Hi, thank you for providing an excellent library.  I noticed that the read_edf also needs to be changed to int32 on line 365 of highlevel.py for the anonymize_edf function to work.

skjerns:
@snbarr feel free to make a PR/issue . 

PS the reference doesnt seem to work, https://github.com/holgern/pyedflib/blob/d7faabf9ef7edc7db0a953b5ba1e19d9f973db1f/pyedflib/highlevel.py#L365

snbarr:
I think there is already an issue for this.  https://github.com/holgern/pyedflib/issues/54  Basically what happens is that read_edf uses int and write_edf expects int32.  I'm not sure why int isn't working, but when I change the line below to int32 it works.

https://github.com/skjerns/pyedflib/blob/ec6ec41513dd6db13479e28d2ed82fbafabe90e8/pyedflib/highlevel.py#L365

https://github.com/holgern/pyedflib/blob/d7faabf9ef7edc7db0a953b5ba1e19d9f973db1f/pyedflib/highlevel.py#L368

PR #65, opened 2020-01-07
Added high-level functions

skjerns:
I thought it might be useful to add some high-level functions for convenience.

I've created the following functions:

```
write_edf(filename, data)
read_edf(filename)

drop_channels(filename, to_drop) # drops channels or keeps only a selected few
anonymize_edf(filename) # remove all personalized data from a header
rename_channels # renames channels in an EDF file

# helper functions
make_header()  # create a header dictionary
make_signal_header() # create a signal_header
```

@holgern  Let me know if you think this could be useful. Then I'll create the tests and integrate the imports into the package.


skjerns:
(note: need to add digitial for non-same sampling frequency)

skjerns:
@holgern do you think it's worth pursuing this?

holgern:
@skjerns Thanks for your work. I would like to have the added dependencies dropped or made optional.  They can be added to the test requirements but should not be required to install the package itself.

skjerns:
@holgern I added the docstrings and tests. Can you see if this can be merges? I'll write some more documentation then.

btw, I would volunteer to take up part of maintenance of the repository, if you are fine with that?

holgern:
Great work, it looks good now. This will simplify the usage of pyedflib.

I think I have to create a new github organization first, before I can add maintainer.

skjerns:
@holgern I think it should work without an organiztion, you just have to click on the settings above and under `manage access` add me as collobator

holgern:
@skjerns You were right, I invited you as collaborator.

Issue #72, opened 2020-03-04
Numpy dependency


MeNsaaH:
Instead of failing on installation if numpy is not found, it'll be better to add it to setup.py `install_requires`

skjerns:
I'm quite surprised, because numpy is in the requirements: 

`install_requires=["numpy>=1.9.1"]`

https://github.com/holgern/pyedflib/blob/0fd2c3e9c164c2881f02ac252ee19ff78c0c338c/setup.py#L287

MeNsaaH:
Why then does it keep failing if numpy is not preinstalled? Is there some numpy functions called during setup?

skjerns:
yes, you're right, there is some code executed before setup using numpy.

not a priority right now, as 95% of Python users probably have numpy installed already.

You're welcome to make a PR that addresses the issue!

MeNsaaH:
Sure

PR #73, opened 2020-03-04
Install Numpy during installation and not fail when not installed

MeNsaaH:
using setuptools `dist.Distribution().fetch_build_eggs`, install numpy if it's not already found installed. This is with respect to #72 earlier opened


skjerns:
looks good, thanks!

PR #75, opened 2020-03-31
small fix

skjerns:
small fix for #57


skjerns:
@snbarr does this fix it for you?

PR #81, opened 2020-05-14
Update to EDFlib 1.17

skjerns:
Pulled most recent version of `edflib`. Now we have subsecond starting sequence accuracy (will be implemented in a second PR)

I've uncommented line 3649, as before, as otherwise we can't create vanilla EDF/BDF files. Why is this line there? Seems counter intuitive.

@holgern : can you have a look? as this introduces some major changes. All tests run without problem, so I assume that the update didn't break anything.


holgern:
I did same modifications to EDFlib, so just updating EDlib will not work.

We need to port changes manually other wise pyedflib will
* lose the option to disable filesize check line 242
* Lose more specific error message, I introduced EDFLIB_FILE_ERRORS_LABEL and others
* Allow channels where ((edfhdr->edfparam[i].dig_max == edfhdr->edfparam[i].dig_min)
* Lose edflib_repair_file_size
* Lose option to write EDF files

We need to port the changes, not just overwrite the c and h file.

skjerns:
@holgern: I was afraid so already. I'll see how far I'll get with restoring the changes.

skjerns:
@holgern I integrated all the changes and also wrote some additional tests, e.g. to check if we're really writing EDF and not EDF+. Everything seems to work now. Can you check? 

Some questions:
- Why retain the ability to write vanilla EDF? EDF+ is backwards-compatible, so all EDF-software should be able to read EDF+ (if no extra-features of EDF+ are used)

Once merged I'll address the following issues:
- read/write sub-second starttime correctly
- read/write patient information for vanilla EDF files correctly

holgern:
Looks good to me, tanks a lot :).

PR #84, opened 2020-07-05
improve docstr of anonymize_edf

sappelhoff:
I added some clarifications to the docstring of `anonymize_edf` (no change in code behavior).

When saving, my editor automatically removed all trailing whitespace in the document (that's why there are many marked lines of no visible changes)


skjerns:
seems fine, no idea why the test fails. seems not to be related to the changes.

PR #89, opened 2020-10-01
Assert that physical_min is different from physical_max

LucaCerina:
I've noticed that if physical_min and physical_max are equal, write_samples fails with error -25. An assertion before writing can help the user in understanding the issue


skjerns:
Thanks for the pull request :) looks good to me.

Two things:
1. Can you write a small unit test? Always good to have those.
2. It might be an idea to include this in the lowlevel, not highlevel, i.e. [edfwriter.py](https://github.com/holgern/pyedflib/blob/master/pyedflib/edfwriter.py) - would you be interested in doing that?

LucaCerina:
Sure, I will add the unit tests and move the assertion at a lower level

skjerns:
thanks alot, looks good to me :)

Issue #92, opened 2020-10-21
write_edf is very slow after a signal has been changed


DavidCEAMS:
Hi,
I'm trying to use the high lvl API to read an EDF, filter a signal and write it back. The problem is that when I do that, the call to write_edf takes several minutes to complete (for a 400MB file).

I'd like to make sure I'm using it correctly.

Here's a sample code:
```
import pyedflib
import numpy as np
from scipy import signal

def find_channel_index(signal_headers, channel_label):
    for i, header in enumerate(signal_headers):
        if header['label'] == channel_label:
            return i
    return None

input_filename = 'subject1.edf'
output_filename = 'subject1_output.edf'
channel = 'EEG Fz'

# Open the EDF file
[signals, signal_headers, file_header] = pyedflib.highlevel.read_edf(input_filename)

# Find the channel to filter
channel_index = find_channel_index(signal_headers, channel)

# Create the filter
lf = 8
hf = 12
order = 100
sample_rate = signal_headers[channel_index]['sample_rate']
taps = signal.firwin(order, [lf, hf], pass_zero='bandpass', fs=sample_rate)

# Filter the signal
signals[channel_index] = signal.filtfilt(taps, 1, signals[channel_index])

# Write it back
pyedflib.highlevel.write_edf(output_filename, signals, signal_headers, file_header, file_type=pyedflib.FILETYPE_EDF)
```
Is this the proper way of doing it?

Thanks

skjerns:
Can you upload the sample file to wetransfer (or dropboxlink etc) and send it to tm2v0apq1o2@temp.mailbox.org ?

holgern:
The only way to speed up writing huge data files is to use setDatarecordDuration

f_writer.setDatarecordDuration(int(100000 * datarecord_duration))
When using this, each sample rate needs to be multiplied by it:

ch_dict['sample_rate'] = ch_dict['sample_rate'] * datarecord_duration

Maybe this would be a good addition to the high-level code.

skjerns:
Ah, I didn't know about that. However, that means if the `datarecord_duration` is not a divisor of the length of the signal (in seconds), it will append zeros which I think it not favorable. E.g. writing a signal with length 150 seconds with `datarecord_duration=60` will write 150 seconds + 30 seconds blank.

I'll implement it such that it will only choose an appropriate `datarecord_duration`.

holgern:
[...]

PR #93, opened 2020-10-22
Automatically set blocksize // speed up writing

skjerns:
I've implemented automatic setting of the blocksize, it will default to the highest `int` divisor below 60. As far as I can see `datarecord_duration` is in fact a float and could be set to any value, however, I'm a bit afraid that this will introduce side effects and rounding errors that result in loss of samples.

I did not want to include a manual trim while loading, as other EDF loading software does not implement that. However, the `block_size` parameter can be set manually if people do not mind having padded zeros at the end.

related: #92 

@holgern can you check, approve and possibly make a new release?

### DIFF
--- a/pyedflib/tests/test_highlevel.py
+++ b/pyedflib/tests/test_highlevel.py
@@ -56,7 +56,7 @@ class TestHighLevel(unittest.TestCase):
         
         header = highlevel.make_header(technician='tech', recording_additional='r_add',
                                                 patientname='name', patient_additional='p_add',
-                                                patientcode='42', equipment='eeg', admincode='420',
+                                                patientcode='42', equipment='eeg', admincode='120',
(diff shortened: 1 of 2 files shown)


pyedflib/tests/test_highlevel.py
```diff
@@ -56,7 +56,7 @@ def test_read_write_edf(self):
         
         header = highlevel.make_header(technician='tech', recording_additional='r_add',
                                                 patientname='name', patient_additional='p_add',
-                                                patientcode='42', equipment='eeg', admincode='420',
+                                                patientcode='42', equipment='eeg', admincode='120',
```
skjerns, line 59: after all we're doing science, not 420 ;-)

holgern:
Sorry for the delay.

PR #114, opened 2021-02-26
Be less strict about prefilter fields

wmvanvliet:
I'm trying to load EDF file from this dataset: https://physionet.org/content/eegmmidb
In these files, prefilter settings are present in both the normal fields and also in the EDF+ annotations.
For some reason, it was decided that pyedflib should flat out refuse to read such files?
I think that is a bit harsch.

### DIFF
--- a/pyedflib/_extensions/c/edflib.c
+++ b/pyedflib/_extensions/c/edflib.c
@@ -41,7 +41,6 @@
 
 #include "edflib.h"
 
-
 #define EDFLIB_VERSION  (117)
 #define EDFLIB_MAXFILES  (64)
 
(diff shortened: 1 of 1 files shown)


skjerns:
I understand the annoyance with being so strict about keeping to the EDF specifications. The library we wrap `edflib` is very strict with regards to these. Even though I would be kind of fine to just read these incorrectly formatted files anyway, I also see why being lose with standards will not help in the long run.

In this particular case: What would happen if prefilter fields have been set to both positions? Which one would be reported?

wmvanvliet:
In the original code, it would crash. With this PR, the first would be taken.

skjerns:
fixes #120

PR #116, opened 2021-03-23
fixed wrong variable use in rename_channel function

salman2135:
HI,
I found a bug while using rename_channel function. I believe wrong variable was used to call read_edf function. I changed it to following:

`    for ch_nr in tqdm(range(len(channels))):
        signal, signal_header, _ = read_edf(edf_file, digital=True,
                                            ch_nrs=ch_nr, verbose=False)`
Hope it helps!


skjerns:
thanks, that's a bug indeed! I have it on my list since a long time to write some unittests for these functions..

PR #118, opened 2021-05-03
Decimal sample rate compatibility

gcathelain:
No error is raised when reading or writing with sample_rate with decimals.

Block size must be -1 when using the pyedflib.highlevel.edfwriter() function.


skjerns:
@gcathelain so this produces quite a few errors, I don't think we can merge it like that. We need to find a solution that both

1. conserves `int` sampling rates
2. allows decimal sampling rates

do you plan on proposing a solution for that?

pyedflib/highlevel.py
```diff
@@ -327,19 +327,19 @@ def read_edf(edf_file, ch_nrs=None, ch_names=None, digital=False, verbose=False)
 
     with pyedflib.EdfReader(edf_file) as f:
         # see which channels we want to load
-        available_chs = [ch.upper() for ch in f.getSignalLabels()]
+        available_chs = f.getSignalLabels()
```
skjerns, line 330: why is this changed?

pyedflib/highlevel.py
```diff
@@ -327,19 +327,19 @@ def read_edf(edf_file, ch_nrs=None, ch_names=None, digital=False, verbose=False)
...
         n_chrs = f.signals_in_file
 
         # find out which number corresponds to which channel
         if ch_names is not None:
             ch_nrs = []
             for ch in ch_names:
-                if not ch.upper() in available_chs:
+                if not ch in available_chs:
```
skjerns, line 337: why is this changed? seems unrelated to the PR

pyedflib/highlevel.py
gcathelain: It was to simplify, but I agree it is unrelated to the PR.

pyedflib/highlevel.py
gcathelain: It was to simplify, but I agree it is unrelated to the PR.

skjerns:
Mhh now the tests don't run anymore, I think this needs some further adapting before it works. However, I currently do not have time to look into this. 

Additionally, it would be important to also check the newly implemented functionality using unit tests

skjerns:
This is superseeded by #159

PR #121, opened 2021-05-05
Fix sample frequencies for record duration != 1

### LINKED ISSUE #117: EdfWriter writes incorrect sample frequencies for record lengths != 1
I'm working with EDF files that have record lengths other than 1 second and am trying to generate dummy files for testing, and I think I've encountered a bug with how `EdfWriter` handles sample rate vs sample frequency. I'm writing a file with a sample rate of 256, a record length of 2, and a total of 6 minutes of data (example code below). I expect the generated EDF file to have 180 records with 512 samples per record, but the actual file has 360 records with 256 samples per record.

I think I found the source of the problem with regards to calculating sample frequency: in `EdfWriter.update_header`, channel sample rates are passed to `set_samplefrequency`, but inside that function the argument is named `samplefrequency` and treated as such. My understanding is that sample rate represents samples-per-second, and sample frequency represents samples-per-record. So treating them interchangeably would result in incorrect numbers for record lengths != 1. I saw #111 and suspect that this is related.

I haven't found where number of records is calculated yet, but I imagine the issue is similar for that header attribute as well.

Example code:
```
import numpy as np
from pyedflib import EdfWriter, FILETYPE_EDF
from datetime import datetime

MILLISECONDS = 1000
TEN_MICROSECONDS = MILLISECONDS * 100
SIX_MINUTES = 360

filename = 'edf_file.edf'
sample_rate = 256
samples_per_record = 512  # For a 2 sec record length
record_length = (samples_per_record / sample_rate) * TEN_MICROSECONDS
channel_names = ['ch1', 'ch2', 'ch3', 'ch4', 'ch5']
signals = np.random.rand(len(channel_names), sample_rate * SIX_MINUTES) * 200
start_date = datetime(2017, 5, 24, 15, 53, 49)

writer = EdfWriter(filename, len(channel_names), file_type=FILETYPE_EDF)
writer.setPatientCode('')
writer.setRecordingAdditional('')
writer.setDatarecordDuration(record_length)
writer.setStartdatetime(start_date)
writer.setSignalHeaders([{
    'label': channel,
    'sample_rate': sample_rate,
    'dimension': 'uV',
    'physical_min': -200.0,
    'physical_max': 200.0,
    'digital_min': -32768,
    'digital_max': 32767,
    'transducer': f'EEG {channel}',
    'prefilter': ''
} for channel in channel_names])

writer.writeSamples(signals)
```

Let me know if my assessment is correct (I'm new to working with EDF files, so I'm still getting across how they're structured), and I'd be happy to open a PR.

cfranklin11:
Resolves #117 

The terms 'sample_rate' and 'sample_frequency' were used
interchangeably in the codebase, which is valid as long as the
record duration is 1 second, but they stop being the same when
the record duration is any other value. I followed the meaning
established in the documentation, where 'sample_frequency' is
taken to represent the number of samples per signal per record,
and 'sample_rate' is the number samples per signal per second.

This represents a breaking change to pyedflib's API, as it changes
the name of various EdfWriter method parameters and keys
in the signal dictionaries returned by EdfReader methods.
However, it conforms to the terminology used in the EDF spec
(that the header displays the number of samples in each data
record) as well as edflib function names.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,13 +90,13 @@ Additionally functionality as anonymizing, dropping or renaming channels can be 
     # write an edf file
     signals = np.random.rand(5, 256*300)*200 # 5 minutes of random signal
     channel_names = ['ch1', 'ch2', 'ch3', 'ch4', 'ch5']
-    signal_headers = highlevel.make_signal_headers(channel_names, sample_rate=256)
+    signal_headers = highlevel.make_signal_headers(channel_names, sample_frequency=256)
     header = highlevel.make_header(patientname='patient_x', gender='Female')
     highlevel.write_edf('edf_file.edf', signals, signal_headers, header)
-	    
+
     # read an edf file
     signals, signal_headers, header = highlevel.read_edf('edf_file.edf')
-    print(signal_headers[0]['sample_rate']) # prints 256
+    print(signal_headers[0]['sample_frequency']) # prints 256
 
     # drop a channel from the file or anonymize edf
     highlevel.drop_channels('edf_file.edf', to_drop=['ch2', 'ch4'])
--- a/demo/writeEDFFile.py
+++ b/demo/writeEDFFile.py
@@ -32,7 +32,7 @@ if __name__ == '__main__':
     channel_info = []
     data_list = []
 
-    ch_dict = {'label': 'squarewave', 'dimension': 'uV', 'sample_rate': 200, 'physical_max': 100, 'physical_min': -100, 'digital_max': 32767, 'digital_min': -32768, 'transducer': '', 'prefilter':''}
+    ch_dict = {'label': 'squarewave', 'dimension': 'uV', 'sample_frequency': 200, 'physical_max': 100, 'physical_min': -100, 'digital_max': 32767, 'digital_min': -32768, 'transducer': '', 'prefilter':''}
     channel_info.append(ch_dict)
     time = np.linspace(0, file_duration, file_duration*200)
     xtemp = np.sin(2*np.pi*0.1*time)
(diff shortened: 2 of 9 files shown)


skjerns:
Thank you very much for this great contribution! 

I am a bit puzzled why the build failed, it seems to be rather a build config problem. However, I am unable to trigger a new build manually, @holgern do you know something about this?

holgern:
The builds failed due to a timeout, as they tried to build numpy from scratch (which is strange).
So the failed build has nothing to do with the changes within this PR.

holgern:
I agree for the change of sample_rate to sample_frequency, but this could lead to problems.

We need a deprecation warning whenever the old parameter is used.
@cfranklin11 Could you add the old parameter and give a deprecation whenever they are used?

holgern:
We can then remove the old names in the next next release.

cfranklin11:
Looking into having both `sample_rate` and `sample_frequency` work in parallel, and it's fine in most places, but I'm having some trouble with `EdfWriter#set_header`, because we have to decide which channel attribute to use when calling `set_samplefrequency`. Initially, I was going to check for the default value of 100 used in `EdfWriter.__init__`, but then I saw that the default is 256 when using the `highlevel` interface.

My idea is to make the default for `sample_frequency` `None` to start, so we can use that value if it's present and fall back to `sample_rate` if it's not. Then, when we remove `sample_rate` in a future release, `sample_frequency` can adopt the existing default values. Let me know if this sounds reasonable.

skjerns:
@cfranklin11 (sorry for the late reply)

That sounds good! 
On second thought, I think it doesn't make sense anyway to assume a default sample frequency/rate. The sample rate is almost as important to the signal as the signal itself, it is not optional. I'd be fine with both, using 'None' (to not change compatibility) or making it an non-keyword.

skjerns:
Some tests need some minor fixing now that the defaults have changed (somehow I can't push to your branch, else I'd fix it for you)

Also I fixed the Travis timeouts in #127 , so if you fetch&merge upstream they will finally work again :)

skjerns:
great work! Thanks alot for the effort @cfranklin11 :)

Issue #123, opened 2021-06-03
EdfWriter setSignalHeader fails if transducer/prefilter not specified 


BlakeJC94:
I'm currently using this package to write EDF files where different channels have different sample rates and I came across a bug. Workarounds and fixes should be straightforward, (happy to make those changes if needed).

Tried using the function `setSignalHeader` as instructed in the comments, but the docs don't mention `transducer` and `prefilter` are also required in the `channel_info` dictionary (probably also happens for `setSignalHeaders` as well).

Simple script that replicates error:
```python
from pyedflib import EdfWriter
# parameters
n_sec = 3
ch_grps = [[0, 1, 2], [3, 4], [5]]
s_rate = [128, 64, 256]
filename = 'test.edf'
# generate test signal
signals = list()
for idx_grp, ch_grp in enumerate(ch_grps):
    for ch in ch_grp:
        signals.append(list(range(n_sec*s_rate[idx_grp])))
# write test file
n_ch = len(signals)
edf = EdfWriter(filename, n_ch)
for idx_grp, ch_grp in enumerate(ch_grps):
    for ch in ch_grp:
        channel_info = {'label': f'test_ch_{ch}',
                        'dimension': 'uV',
                        'sample_rate': s_rate[idx_grp],
                        'physical_max': 200.0,
                        'physical_min': -200.0,
                        'digital_max': 32768,
                        'digital_min': -32767}
        edf.setSignalHeader(ch, channel_info)
edf.close()
```

Error produced:
```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-33-3421a72324e0> in <module>
     39                         'digital_max': 32768,
     40                         'digital_min': -32767}
---> 41         edf.setSignalHeader(ch, channel_info)
     42 edf.close()

~/.local/lib/python3.8/site-packages/pyedflib/edfwriter.py in setSignalHeader(self, edfsignal, channel_info)
    237             raise ChannelDoesNotExist(edfsignal)
    238         self.channels[edfsignal] = channel_info
--> 239         self.update_header()
    240 
    241     def setSignalHeaders(self, signalHeaders):

~/.local/lib/python3.8/site-packages/pyedflib/edfwriter.py in update_header(self)
    200             set_label(self.handle, i, du(self.channels[i]['label']))
    201             set_physical_dimension(self.handle, i, du(self.channels[i]['dimension']))
[...]

PR #124, opened 2021-06-03
updated edfWriter:setSignalHeaders to update default dicts instead of overwriting

BlakeJC94:
PR for #123 
Implements suggested changes in post, all tests confirmed passing.


skjerns:
Looks good!

Thank you very much :)

PR #125, opened 2021-06-05
non-empty data check in edfwriter to prevent unknown OS error

BlakeJC94:
Ran into an issue on Ubuntu and Windows that would cause edfWriter to wither hang or crash and return `IOError('Unknown error while calling blockWriteSamples')` (edfWriter.py:741). Investigated and found this occurs if `len(data_list) == 0` in `edfWriter.writeSamples()`.

Added another conditional for this on line 693 to save future users effort in debugging.

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -690,8 +690,11 @@ class EdfWriter(object):
         """
 
 
-        if (len(data_list) != len(self.channels)) or (len(data_list) == 0):
-            raise WrongInputSize(len(data_list))
+        if (len(data_list) == 0:
+            raise WrongInputSize('Data list is empty') 
+        if (len(data_list) != len(self.channels)):
+            raise WrongInputSize('Number of channels ({}) \
+             unequal to length of data ({})'.format(len(self.channels), len(data_list)))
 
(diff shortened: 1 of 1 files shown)


skjerns:
Thanks alot!

Maybe we can split up the errors in two different errors and include a more detailed error message?

pyedflib/edfwriter.py
```diff
@@ -690,7 +690,7 @@ def writeSamples(self, data_list, digital = False):
         """
 
 
-        if (len(data_list) != len(self.channels)):
+        if (len(data_list) != len(self.channels)) or (len(data_list) == 0):
             raise WrongInputSize(len(data_list))
```
skjerns, line 694:
```suggestion
        if (len(data_list) == 0:
            raise WrongInputSize('Data list is empty') 
        if (len(data_list) != len(self.channels)):
            raise WrongInputSize('Number of channels ({}) \
             unequal to length of data ({})'.format(len(self.channels), len(data_list)))
```

skjerns (changes requested):

BlakeJC94 (commented):
Looks good! It's a minor thing to change, but hopefully this improves usability

skjerns (approved):

PR #126, opened 2021-06-05
Fix annotation bytestring

### LINKED ISSUE #119: EDFWriter raises error when writing annotations
Hi, I am writing an edf file with `highlevel.write_edf` using the output of `highlevel.read_edf`. My example file is attached. The error is raised at line 736 of `edfwriter.py` for `if duration_in_seconds >= 0`: `E           TypeError: '>=' not supported between instances of 'bytes' and 'int'`.
This is because in the annotations of the header, the annotation event's `duration_in_seconds` is read as a bytes type. I tried to convert that to `float`, it worked and wrote out an edf file. This file once read in again has `duration_in_seconds` of bytes type. Could you fix either the reader or writer?
[out.edf.zip](https://github.com/holgern/pyedflib/files/6419525/out.edf.zip)

Thanks for your attention.

skjerns:
fixes #119 

I changed the way annotations are loaded slightly, and there is an explicit type conversion happening now

I also use the internal function `readAnnotations` instead of the Cython `read_annotations`, which already has all the type conversions that are needed. This should have been the way anyway.

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -769,6 +769,10 @@ class EdfWriter(object):
         """
         if self.file_type in [FILETYPE_EDF, FILETYPE_BDF]:
             raise TypeError('Trying to write annotation to EDF/BDF, must use EDF+/BDF+')
+
+        if isinstance(duration_in_seconds, bytes):
+            duration_in_seconds = float(duration_in_seconds)
+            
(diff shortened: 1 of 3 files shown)


skjerns:
@holgern could you check the Travis CI and see why the tests timeout and try to fix it? 
I don't seem to have the access rights.

PR #137, opened 2021-07-01
improved speedup write with F-contiguous arrays

LucaCerina:
Referencing discussion in #92 all Fortran-contiguous arrays are transformed before writing. This vastly improves write speed for this kind of arrays.

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -817,14 +817,14 @@ class EdfWriter(object):
             del dataRecord
             dataRecord = np.array([], dtype=np.int32 if digital else None)
             for i in np.arange(len(data_list)):
-                dataRecord = np.append(dataRecord,data_list[i].ravel()[int(ind[i]):int(ind[i]+sampleFrequencies[i])])
+                dataRecord = np.append(dataRecord, data_list[i][int(ind[i]):int(ind[i]+sampleFrequencies[i])])
                 ind[i] += sampleFrequencies[i]
             if digital:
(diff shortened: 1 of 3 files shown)


pyedflib/edfwriter.py
```diff
@@ -817,14 +817,14 @@ def writeSamples(self, data_list, digital = False):
             del dataRecord
             dataRecord = np.array([], dtype=np.int32 if digital else None)
             for i in np.arange(len(data_list)):
-                dataRecord = np.append(dataRecord,data_list[i].ravel()[int(ind[i]):int(ind[i]+sampleFrequencies[i])])
+                dataRecord = np.append(dataRecord, data_list[i][int(ind[i]):int(ind[i]+sampleFrequencies[i])])
```
skjerns, line 820: now you removed the `ravel`, was this intentional? right now the fix would onyl apply to the highlevel-interface (which is okay I guess)

skjerns:
also in `edfwriter.py` there is the line in 787-791 :) you can just copy&paste the segment from before.

LucaCerina:
You are right, I forgot people may be calling edfwriter directly. As a solution I moved the check at edfwriter level instead of highlevel. Once its done before writing samples, we don't need the other `ravel` calls because 'data_list[i]' is already flattened and C-contiguous

skjerns:
thanks :) looks good now!

skjerns:
oops appareantly the "all checks passed" was a display error and now some builds fail. I'll check it out and fix it if necessary. Should not be a major bug though

PR #140, opened 2021-07-14
fix badges, add pypi badge

sappelhoff:
1. Adds "alt" text to each badge that shows up when the SVG cannot be loaded
2. removes one of the two conda badges, which was simply pointing to the conda-forge repository (all packages therein):  https://conda.anaconda.org/conda-forge  
3. replaces the pyedflib-specific conda-forge badge with one that explicitly says that this package is on conda-forge
4. Adds a PyPI badge next to the conda badge

Before:

image

After:

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,28 +4,36 @@ pyEDFlib
 .. contents::
 
 .. image:: https://codecov.io/gh/holgern/pyedflib/branch/master/graph/badge.svg
-  :target: https://codecov.io/gh/holgern/pyedflib
+   :target: https://codecov.io/gh/holgern/pyedflib
+   :alt: Test Coverage
 
 .. image:: https://travis-ci.com/holgern/pyedflib.svg?branch=master
-    :target: https://travis-ci.com/holgern/pyedflib
+   :target: https://travis-ci.com/holgern/pyedflib
+   :alt: Travis Build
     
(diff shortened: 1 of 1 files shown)


skjerns:
Great add! Thanks a lot for the contribution. 

Do you know why the coverage badge doesn't work?

sappelhoff:
> Do you know why the coverage badge doesn't work?

I took a brief look at the different CIs you have running, but I wasn't sure how welcome meddling would be so I stopped again :-) 

I can take another look if you want.

sappelhoff:
Also, the code quality badge seems broken:

image

skjerns:
Ah okay thanks again, and also for looking into the coverage badge. 

I don't think I can fix the Travis as I am not the owner of the repo, I guess @holgern  would need to do that. But I'd say that's low prio

sappelhoff:
> But I'd say that's low prio

you mean because running CircleCI and appveyor is sufficient? :thinking: 

Because the whole Travis CI is not running, it's not just the coverage reports that are missing.

skjerns:
Ah that could be! Sorry I'm on mobile right now, can't easily check it. I'll look into it next week :)

skjerns:
looks good, let's try to fix the coverage in #141

PR #141, opened 2021-07-14
try to fix codecov upload -> TRAVIS authorization problem

sappelhoff:
as discussed in #140 

not sure if it works, but worth a try.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -177,16 +177,12 @@ script:
     else
         CFLAGS="--coverage" python setup.py build --build-lib build/lib/ --build-temp build/tmp/
         CFLAGS="--coverage" pip install -e . -v
-        pushd demo
-        pytest --pyargs pyedflib --cov=pyedflib --cov-config=../.coveragerc
-        cp .coverage ..
-        popd
+        python -m pytest --pyargs pyedflib --cov=pyedflib --cov-config=.coveragerc --cov-report=xml --verbose
     fi
 
 after_success:
   - |
     if [ "${MANYLINUX}" == "0" ]; then
-        codecov
         # Travis CI has old gcov, needs object dir explicitly specified
         gcov --object-directory build/tmp/pyedflib/_extensions/c/ build/tmp/pyedflib/_extensions/c/*.o
         bash <(curl -s https://codecov.io/bash)


sappelhoff:
@skjerns the problem is that Travis is not building at all: https://travis-ci.com/github/holgern/pyedflib/requests

Some authorization problem:

image

sappelhoff:
relevant: https://stackoverflow.com/a/41327657/5201771

.travis.yml
```diff
@@ -177,16 +177,12 @@ script:
...
-        pytest --pyargs pyedflib --cov=pyedflib --cov-config=../.coveragerc
-        cp .coverage ..
-        popd
+        python -m pytest --pyargs pyedflib --cov=pyedflib --cov-config=.coveragerc --cov-report=xml --verbose
     fi
 
 after_success:
   - |
     if [ "${MANYLINUX}" == "0" ]; then
-        codecov
```
skjerns, line 189: sure this is correct?

sappelhoff: yes, `codecov` only gathers the coverage report and submits it to codecov:

> Find coverage reports for all the languages below, gather them and submit them to Codecov.

source: https://github.com/codecov/codecov-python

The same is achieved by the small bash snippet a few lines below: 

> bash <(curl -s https://codecov.io/bash)

Last time I checked, that bash snippet was the recommended mode of uploading reports to codecov.

Removing the python `codecov` package here just prevents duplication.

skjerns:
@holgern can you check the authorization on travis/coverage?

however, it's a bit strange as the regular builds work fine

sappelhoff:
> however, it's a bit strange as the regular builds work fine

but the last Travis CI builds are more than a month past - despite several merges (other than PRs) in between :thinking: 

Or are both of us looking at different data? --> https://travis-ci.com/github/holgern/pyedflib/builds

.travis.yml
skjerns: thanks!

holgern:
@skjerns I checked the authorization and they look fine. Should we change the coverage tool?, maybe there is a better one or the coverage should not be uploaded from travis.

skjerns:
yes, that seems like a good idea. currently there are three CI tools active anyway (travis, circle, AppVeyor ?), I guess one or two should be sufficient.

I guess the coverage should be able to upload from either of those

skjerns:
Alternatively we could just use github actions for running tests?

+ best integration and no authorization problems
- less verbose interface and might be a bit more cumbersome if stuff fails

example from one of my other repos 
https://github.com/Unisens/pyunisens/blob/master/.github/workflows/Tests.yml

skjerns:
I created a workflow for codecov and it works now. 

However I'm not sure if the Cython code was validated before, and if that's now missing?

I also removed all traces of travis

sappelhoff:
nice, thanks for fixing this @skjerns

PR #150, opened 2021-11-20
Support a system-wide edflib (fix #149)

### LINKED ISSUE #149: Support building with a system copy of edflib?
For some users, and particularly for Linux distribution packagers, it is beneficial to be able to build `pyedflib` using a system-wide copy of `edflib`. [This Fedora Linux wiki page](https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries) describes some of the reasons this can be desirable. It would be helpful if there were some mechanism to support this without patching `setup.py`.

Thanks for considering this request. I will follow up with a PR offering one possible implementation.

musicinmybrain:
Adds an environment variable `SYSTEM_EDFLIB`. When it is set empty or to `1`, the bundled EDFlib sources are not compiled, and a system-wide `edf` library is linked instead.

### DIFF
--- a/pyedflib/_extensions/c_edf.pxd
+++ b/pyedflib/_extensions/c_edf.pxd
@@ -6,7 +6,7 @@
 # get constants
 include "edf.pxi"
 
-cdef extern from "c/edflib.h":
+cdef extern from "edflib.h":
     int edf_set_patientcode(int, char *)
     int edfwrite_annotation_latin1(int, long long int, long long int, char *)
     int edfwrite_annotation_utf8(int, long long int, long long int, char *)
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,10 @@ except ImportError:
                "version of PyEDFlib")
         raise RuntimeError(msg)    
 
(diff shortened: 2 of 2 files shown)


skjerns:
Thanks for the PR, that looks interesting indeed. I'll have a look at it the coming days. As it is purely additional and does not infer with the commonly used install mode, it should be fine to merge.

musicinmybrain:
I see now that the bundled EDFlib is actually a fork, with at least some added error codes and an extra size-check parameter to `edfopen_file_readonly`. As long as `pyedflib` needs to fork EDFlib, this PR isn’t actually useful.

skjerns:
> I see now that the bundled EDFlib is actually a fork, with at least some added error codes and an extra size-check parameter to `edfopen_file_readonly`. As long as `pyedflib` needs to fork EDFlib, this PR isn’t actually useful.

Yes, that is true. However, we could think about making some of the changes optional, I don't think all of them are extremely important. On the other side, it should be no problem to recompile the source for installing the lib.

PR #153, opened 2021-12-10
Remove limit on numpy version

jcbsv:
There is no guarantee that a specified numpy version has a wheel package for any given python version. By relaxing the numpy version limit, pip will fetch the available numpy wheel package, instead of trying to build the specific numpy version (which may fail to complete).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools",
     "wheel",
-	"numpy <= 1.19.3",
+	"numpy",
     "cython"
 ]
 build-backend = "setuptools.build_meta"


skjerns:
perfect, thanks!

closes #152

cbrnr:
Any plans when the next release will be out? Currently it is not possible to just `pip install pyEDFlib` on Python 3.10.

skjerns:
@holgern could you make a new release for python 3.10? :)

flyingcow622:
How exactly do you go about relaxing the numpy version limit? sorry, I'm new to python.

skjerns:
I think if you download the current release it should be already without version limit. What error are you getting? Are you on Python 3.10?

PR #154, opened 2022-01-04
Build Python 3.10 wheels for aarch64

cbrnr:
As mentioned in #153, it would be great if a new release with Python 3.10 wheels could be made. This PR adds wheels for aarch64.


cbrnr:
Let me know if more changes are required, e.g. add a Python 3.10 job to CircleCI. It would also be great to see if these changes work, but e.g. wheels are only built for tags it seems.

skjerns:
Ah thanks! I was not even aware this script had been added.

Issue #111, opened 2021-02-16
Write EDF with sample rate < 1


skjerns:
Reading and writing sample rates generally seems to work, however, there is still a bug with writing sample rates below 1, e.g. 0.5

```Python
signals = np.random.randint(-2048, 2048, [3, 256*60])
highlevel.write_edf_quick('test.edf', signals.astype(np.int32), sfreq=1/3, digital=True)
signals2, _, _ = highlevel.read_edf('test.edf', digital=True, verbose=True)
np.testing.assert_allclose(signals, signals2)
```

skjerns:
should be fixed by #159

PR #159, opened 2022-01-18
fix writing of decimal sample frequency

### LINKED ISSUE #148: "sample_frequency" is a misleading name for the number of values in a signal
Hi, since v0.1.23 the `getSampleFrequency` method has changed the value it returns from the actual sample rate (aka the number of samples per second) to the number of samples in the signal.

This is also documented in code. I don't know the reason why it changed but there probably was a legit reason.

Now when I used the library I stumbled upon this and tracked down the value coming from `getSampleFrequency` to this line: https://github.com/holgern/pyedflib/blob/master/pyedflib/_extensions/_pyedflib.pyx#L358 which says:
```
            return <double>self.hdr.signalparam[channel].smp_in_datarecord
```

So the actual C edflib interface calls this not sample frequency but `smp_in_datarecord`, which is a much more appropriate name IMO. I really struggled with my code and thought: Well I know I have a sample rate of 500 Hz in my data, but now the file says it is 5000 (my signals are 10 seconds), am I producing the wrong files? (sidenote, I only read BDF files with pyedflib, I'm not writing them with this lib)

This confusion IMO comes solely from the name. I'm happy to calculate the frequency (or rate or however you would like to call it) myself, but I'm 100% sure that someone else later in the code will see my calculation and say "hey there is this getSampleFrequency method provided out of the box, let's use this".

What do you think about changing the name of the attribute altogether and removing any ambiguities?

skjerns:
In this PR I'm fixing some mistakes that have been made before, and include some clarifications w.r.t to `sampling_frequency` and `sample_rate`

**Summary**:
* `sample_frequency` is now the main term being used and denoted the samples in Hz (samples per second)
* we hide/discourage the use of `sample_rate`. To prevent confusion, for now we pretend `sample_frequency` and `sample_rate` are the same term towards the user.
* the use of `record_duration` was entirely wrong. Fixed that now.

**Terms**:
`smp_per_record`: tells us how many samples there are in each record.
`record_duration`: denotes the time window of one record. If all sampling frequencies are ints, this can simply be 1.
`sample_frequency`: is calculated by dividing smp_per_record/record_duration. If we write a file, we have to do the reverse, and find an optimal record_duration such that can actually represent the wished sample_frequency.

**Problem**:
Sampling frequency (samples per seconds in Hz) is not explicitly denoted in EDF. It is calculated by `smp_per_record` (samples, int) / `record_duration` (seconds, float). Previously it was just assumed that `sample_rate` is equivalent to `sample_frequency`, which is not the case, as `sample_rate` is taken directly from `smp_per_record`. However, when the `record_duration` is unequal to 1, this is not the same as the sampling frequency.
Similarly, if we want to save a file with decimal `sampling_frequency`, e.g. 0.5 Hz, we need to make the equation `0.5=smp_per_record/record_duration` work such that `smp_per_record` is an int. That means we need to look for a `record_duration` that matches the `smp_per_record` for all channels in the EDF. I posted a simplified solution that should work in most cases (just try out `record_duration` for 1s to 60s in 1s steps), however, we can think of some more mathematical solution as well later.

**Proposition**
I think nobody cares how many samples are stored within a record_duration. These are technical details. People want to set a sampling frequency. They do not care how this is realized and stored within the file. Therefore I opt to remove any setting of `record_duration` and ``

closes #148  #111

This is a rather large commit, and I would appreciate if someone would take a close look at it @holgern


pyedflib/edfreader.py
```diff
@@ -34,35 +34,36 @@ def _debug_parse_header(filename):
...
+        dmax = [f.read(8).decode() for i in range(nsigs)]
+        prefilter = [f.read(80).decode().strip() for i in range(nsigs)]
+        n_samples = [f.read(8).decode() for i in range(nsigs)]
+        reserved = [f.read(32).decode() for i in range(nsigs)]
+    _ = zip(label, transducer, dimension, pmin, pmax, dmin, dmax, prefilter, n_samples, reserved)
+    values = locals().copy()
+    fields = ['label', 'transducer', 'dimension', 'pmin', 'pmax', 'dmin', 'dmax', 'prefilter', 'n_samples', 'reserved']
+    sheaders = [{field:values[field][i] for field in fields} for i in range(nsigs)]
+    print('\n##### Signal Headers')
+    print(json.dumps(sheaders, indent=2))
```
skjerns, line 66: unrelated, just ignore, just for debugging

pyedflib/edfwriter.py
```diff
@@ -14,7 +14,7 @@
 from ._extensions._pyedflib import open_file_writeonly, set_physical_maximum, set_patient_additional, set_digital_maximum
 from ._extensions._pyedflib import set_birthdate, set_digital_minimum, set_technician, set_recording_additional, set_patientname
 from ._extensions._pyedflib import set_patientcode, set_equipment, set_admincode, set_gender, set_datarecord_duration, set_number_of_annotation_signals
-from ._extensions._pyedflib import set_startdatetime, set_starttime_subsecond, set_samplefrequency, set_physical_minimum, set_label, set_physical_dimension
+from ._extensions._pyedflib import set_startdatetime, set_starttime_subsecond, set_samples_per_record, set_physical_minimum, set_label, set_physical_dimension
```
skjerns, line 17: renamed to accurately reflect what the function actually does.

pyedflib/edfwriter.py
```diff
@@ -448,20 +453,21 @@ def setGender(self, gender):
...
+        lead to unexpected side-effects in the sample frequency calculations.
+        
         Returns 0 on success, otherwise -1.
 
         Parameters
         ----------
-        duration : integer
-            Sets the datarecord duration in units of 10 microSeconds
+        record_duration : integer
+            Sets the datarecord duration in units of seconds
```
skjerns, line 470: This was wrong. We insert the `record_duration` in seconds, not 10mS (this is how it is saved later inside the edf, but not how we indicate it here)

pyedflib/highlevel.py
```diff
@@ -420,11 +420,6 @@ def write_edf(edf_file, signals, signal_headers, header=None, digital=False,
     file_type: int, optional
         choose file_type for saving.
         EDF = 0, EDF+ = 1, BDF = 2, BDF+ = 3, automatic from extension = -1
-    block_size : int
-        set the block size for writing. Should be divisor of signal length
-        in seconds. Higher values mean faster writing speed, but if it
-        is not a divisor of the signal duration, it will append zeros.
-        Can be any value between 1=><=60, -1 will auto-infer the fastest value.
```
skjerns, line 427: the use of block_size was entirely wrong here, my bad. needs to be removed.

pyedflib/highlevel.py
```diff
@@ -496,15 +479,10 @@ def write_edf(edf_file, signals, signal_headers, header=None, digital=False,
...
 
-
-        frequency_key = 'sample_rate' if shead.get('sample_frequency') is None else 'sample_frequency'
-        shead[frequency_key] *= block_size
-
     # get annotations, in format [[timepoint, duration, description], [...]]
     annotations = header.get('annotations', [])
 
     with pyedflib.EdfWriter(edf_file, n_channels=n_channels, file_type=file_type) as f:
-        f.setDatarecordDuration(int(100000 * block_size))
```
skjerns, line 507: completely misunderstood the use of `record_duration`

pyedflib/tests/test_edfreader.py
```diff
@@ -35,7 +38,14 @@ def tearDownClass(cls):
...
                 print(e)
-
+                
+    def tearDown(self):
+        # small hack to close handles in case of tests throwing an exception
+        for obj in gc.get_objects():
+            if isinstance(obj, (EdfWriter, EdfReader)):
+                obj.close()
+                del obj
+
```
skjerns, line 48: if a test fails it will leave the file open. if the same file is accessed in a later test, that will fail again, just due to the file still being open.

added this to close all files that were not closed because a test failed.

pyedflib/tests/test_edfreader.py
```diff
@@ -98,24 +108,24 @@ def test_EdfReader_BDF_datarec_0_5(self):
         np.testing.assert_equal(f.datarecords_in_file, datarecords / datarecord_duration)
         np.testing.assert_equal(f.getFileDuration(), datarecords)
         
-        sample_frequencies = [1000, 800, 500, 975, 999]
+        sample_frequencies = [2000, 1600, 1000, 1950, 1998]
```
skjerns, line 112: these were wrong before. Now they are correct and finally are also what EDFBrowser would display.

pyedflib/tests/test_edfreader.py
```diff
@@ -125,12 +135,12 @@ def test_EdfReader_BDF_datarec_2(self):
         np.testing.assert_equal(f.datarecords_in_file, datarecords / datarecord_duration)
         np.testing.assert_equal(f.getFileDuration(), datarecords)
         
-        sample_frequencies = [1000, 800, 500, 975, 999]
+        sample_frequencies = [500, 400, 250, 487.5, 499.5]
```
skjerns, line 138: these were wrong before. Now they are correct and finally are also what EDFBrowser would display.

pyedflib/tests/test_highlevel.py
```diff
@@ -156,11 +166,14 @@ def test_fortran_write(self):
 
 
     def test_read_write_decimal_sample_frequencies(self):
-        signals = np.random.randint(-2048, 2048, [3, 256*60])
+        # first test with digital signals
+        signals = np.random.randint(-2048, 2048, [3, 256*60+8])
```
skjerns, line 170: +8 as the file is padded by zeros else.

skjerns:
I changed using `@master` to `@v3` and `@v4` as suggested in https://github.com/actions/upload-artifact/issues/41

skjerns:
Went once through all the changes, added some more comments.

Thanks for @gcathelain for reviewing!

I'll merge it soon.

PR #162, opened 2022-01-19
Cleanup CIs

skjerns:
We currently have several CIs runnig with several different trigger points.

I think, running everything through GitHub Actions should be cleanest and work well.

For now I propose:

* Remove CircleCI 
   * seems to give build error anyway
   * @holgern would need to disable it in on their website
* Remove Appveyor
   * takes forever to build
   * I see no advantage over Actions

Only one thing I'm not sure about: Both Actions and AppVeyor should use VS2019 to build on Windows. I'm not sure if they differ here and we get a greater variance of build chains that we cover. 

Additionally:
* Is `tox` actually used? Or can we also remove it? @holgern
* I'd like to include a workflow that automatically triggers on making a release, creates wheels for all platforms and uploads them to PiPy, what do you think?


holgern:
I disabled CircleCI and agree that we do not need that.

holgern:
Visual Studio 2019 is used by appveyer and github actions.
 
 we can change the github action os to windows-2019 in order to garantee that VS2019 is used.
 https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md
 
 I think we can disable also AppVeyoer.
 
 tox was used by travis, so I think it can also be removed.

Issue #156, opened 2022-01-13
ValueError: numpy.ndarray size changed, may indicate binary incompatibility.


YaohuaYang:
Env
macOS Monterey Apple M1 Pro
python 3.9.9
Numpy 1.20.3
pyedflib v0.1.24, v0.1.19

All packages we installed through 
**arch -arm64 pip install packagename==version**

Shows below error while importing pyedflib
**ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject**

Besides, I'm not able to install v0.1.20-v0.1.23 since the numpy version was limited to under 1.20.
And I don't want to upgrade NumPy over 1.20 since I have another dependency.
Can anyone help me on this issue? Thanks!

skjerns:
have you tried the solutions from here?

https://github.com/holgern/pyedflib/issues/94

YaohuaYang:
> have you tried the solutions from here?
> 
> #94
No, it doesn't work. Is it possible for you to compile a osx-arm64 version and upload to conda package manager? It would be supper helpful!
https://anaconda.org/conda-forge/pyedflib

skjerns:
@YaohuaYang is this resolved by #164 ? can you try installing from master?

YaohuaYang:
@skjerns Thanks for suggesting. I solved it with rosetta 2 + conda.

PR #164, opened 2022-01-21
use oldest-supported-numpy to build

hofaflo:
In the v0.1.24 release, pyedflib's cython extension was compiled against Numpy 1.22.1 for Python versions >=3.8 ([see workflow](https://github.com/holgern/pyedflib/runs/4894478892?check_suite_focus=true#step:5:2129)). Trying to import pyedflib in a local environment with a lower minor NumPy version (e.g. 1.21.5) leads to the cython error message shown in #156. 

The discussion on the introduction of an ndarray size change in NumPy 1.20.0 ([link](https://github.com/numpy/numpy/pull/16938)) contains a suggestion to use [`oldest-supported-numpy`](https://pypi.org/project/oldest-supported-numpy/) for building C/cython extensions.

I'm not sure why this error appears at the switch between 1.21 and 1.22 as the size change was introduced in 1.20 and have only tested this on Python 3.10 (Win10, 64bit), but if you and the CI workflows agree, this could be a solution :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools",
     "wheel",
-	"numpy",
+	"oldest-supported-numpy",
     "cython"
 ]
 build-backend = "setuptools.build_meta"


holgern:
Thanks, this seems to be a good solution.

Issue #171, opened 2022-04-13
unused parameter 'mode' in CyEdfReader.open function


lihanhui:
def open(self, file_name, mode='r', annotations_mode=EDFLIB_READ_ALL_ANNOTATIONS, check_file_size=EDFLIB_CHECK_FILE_SIZE):

parameter 'mode' unused!

skjerns:
thanks, well spotted! That seems to have been a leftover/deprecated parameter. I removed it in #172

PR #172, opened 2022-04-17
remove deprecated "mode" parameter

skjerns:
There was a `mode` parameter that seems to have been a leftover of a previous refactor. It was not used indeed as pointed out by #171 so it should be safe to remove.


skjerns:
closes #171

PR #182, opened 2022-09-12
Allow file_type option in drop_channels

aklos:
drop_channels forces the target to EDF+ but it should allow you to choose the file type.

### DIFF
--- a/pyedflib/highlevel.py
+++ b/pyedflib/highlevel.py
@@ -644,7 +644,7 @@ def compare_edf(edf_file1, edf_file2, verbose=False):
 
 
 def drop_channels(edf_source, edf_target=None, to_keep=None, to_drop=None,
-                  verbose=False):
+                  verbose=False, file_type=-1):
     """
     Remove channels from an edf file. Save the file.
     For safety reasons, no source files can be overwritten.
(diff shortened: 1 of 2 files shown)


skjerns:
This should already work - just set `edf_target` to end in `.bdf` and it should automatically be saved with BDF+

However, this change should be fine as well.

aklos:
> This should already work - just set `edf_target` to end in `.bdf` and it should automatically be saved with BDF+
> 
> However, this change should be fine as well.

The issue is that I actually need a plain EDF (no annotations) to work with wfdb, so inferring from file extension isn't enough.

I'll make the requested changes, thanks!

skjerns:
right! so vanilla EDF. Then this check makes sense. I would not have thought that vanilla EDF is still used anywhere.

Does `pyedflib` create valid vanilla EDF files? I do remember there to be some problems

aklos:
They're valid enough that wfdb can process them. I suppose wfdb 4.0.0 now handles EDF/EDF+ without conversion, but a lot of algorithms still rely on using edf2mit which only accepts vanilla EDFs.

I'm very much a layman when it comes to this stuff. Just trying to get some heartbeat classification models working correctly and wfdb is really hard to understand.

pyedflib/highlevel.py
```diff
@@ -441,6 +441,10 @@ def write_edf(edf_file, signals, signal_headers, header=None, digital=False,
         'file_type must be in range -1, 3'
     assert block_size<=60 and block_size>=-1 and block_size!=0, \
         'blocksize must be smaller or equal to 60'
+    if file_type != -1:
```
aklos, line 444: Added a check here in write_edf to make sure file extension matches file_type opt.

skjerns:
looks good, thanks a lot!

PR #184, opened 2022-10-05
Support legacy EDF patient and recording fields

mattbit:
Hi, dealing with old EDF format files (i.e. non EDF+) I noticed that PyEDFlib does not provide access to the ”local recording identification” field.
Although the EDFlib library provides these legacy fields as `patient` and `recording` attributes in the header struct (`edf_hdr_struct`), only `patient` is implemented in PyEDFlib.

In this PR I’ve added the missing `recording` field to the `EdfReader`. For both `patient` and `recording` I've added a `rstrip()` to remove trailing white space since EDFlib always returns the fixed-length string of 80 chars.
I've also added a test to check that these fields are correctly filled when reading a legacy EDF file.

Hope this can be useful, let me know what you think.


skjerns:
Perfect, thank you very much! seems like a useful addition. No idea why it was not there before.

Issue #195, opened 2023-04-13
Add high-level function for truncating to specified start/stop times


raphaelvallat:
Hi,

Oftentimes it is useful to crop/truncate the EDF file to specified start and/or stop times (e.g. time in bed). I suggest adding a new function `pyedflib.highlevel.truncate_edf` (or "crop_edf")

This is an initial working implementation tested on pyedflib 0.1.30

```python
"""Function to crop an EDF file to specified start/end times."""
from pathlib import Path
import pyedflib as pedf
import datetime as dt


def truncate_edf(path_edf, new_start=None, new_stop=None):
    """Truncate an EDF file to desired start/stop times.

    Parameters
    ----------
    path_edf : str
        The path the the EDF file
    new_start : datetime.datetime
        The new start timestamp
    new_stop : datetime.datetime
        The new stop timestamp
    """
    assert isinstance(new_start, (dt.datetime, type(None)))
    assert isinstance(new_stop, (dt.datetime, type(None)))
    path_edf = Path(path_edf)
    assert path_edf.exists(), "File does not exist."

    # Open the original EDF file
    edf = pedf.EdfReader(str(path_edf))
    signals_headers = edf.getSignalHeaders()
    header = edf.getHeader()

    # Define start/stop in samples
    current_start = edf.getStartdatetime()
    if new_start is None:
        new_start = current_start
    start_diff_seconds = (new_start - current_start).total_seconds()
    assert current_start <= new_start

    current_stop = current_start + dt.timedelta(seconds=edf.getFileDuration())
    current_duration = current_stop - current_start
    if new_stop is None:
        new_stop = current_stop
    assert new_stop <= current_stop
    stop_diff_from_start = (new_stop - current_start).total_seconds()

    # Crop each signal
    signals = []
    for i in range(len(edf.getSignalHeaders())):
        sf = edf.getSampleFrequency(i)
        start_idx = int(start_diff_seconds * sf)
        stop_idx = int(stop_diff_from_start * sf)
        signals.append(edf.readSignal(i, start=start_idx, n=stop_idx - start_idx))
    edf.close()

    # Update header startdate and save file
    header["startdate"] = new_start
    outpath = str(path_edf).replace(".edf", "cropped.edf")
    pedf.highlevel.write_edf(outpath, signals, signals_headers, header)

    # Get new EDF start, stop and duration
    edf = pedf.EdfReader(outpath)
    start = edf.getStartdatetime()
[...]

PR #196, opened 2023-04-16
Add highlevel.crop_edf function

raphaelvallat:
As discussed in https://github.com/holgern/pyedflib/issues/195, this adds a new function to crop the EDF file.

I'm still working on the unit tests. The following works like a charm:

```python
from pyedflib import highlevel
from datetime import datetime
new_start = datetime(2011, 4, 4, 12, 58, 0)
new_stop = datetime(2011, 4, 4, 13, 0, 0)
highlevel.crop_edf("data/test_generator.edf", new_start=new_start, new_stop=new_stop, verbose=True)
```

However, when I set both `new_start` and `new_stop` to None, I should get True with `compare_edf(original, cropped)`, but instead got the following error:

image

Do you have any idea why that might be? I'm not changing any of the signal headers 🤔 

PS: I would recommend using [black](https://black.readthedocs.io/en/stable/) for code formatting. I usually set a max line length to 100 instead of the default 80 because I find it more readable. If that's something you'd be interested in let me know and I can submit a second PR after this one.

Thanks,
Raphael


raphaelvallat:
Hi @skjerns!

Haha, that's kind of you but I would not trust myself too much as I've never dive into the `pyedflib` code before 😅 Anyway it's pretty straightforward so hopefully it's not too much work to review.

> Try loading and saving the data with digital=True, the signal should then stay the same

Yep works perfectly, thanks!

I've added unit tests and everything is running smoothly on my machine ✅

pyedflib/tests/test_highlevel.py
```diff
@@ -39,7 +39,7 @@ def tearDownClass(cls):
                 os.remove(os.path.join(data_dir, file))
             except Exception as e:
                 print(e)
-                
+
```
raphaelvallat, line 42: Apologies about the extra blank lines which were automatically added by my VSCode upon saving the file 😬

pyedflib/highlevel.py
```diff
@@ -756,6 +757,81 @@ def anonymize_edf(edf_file, new_file=None,
...
+    current_duration = current_stop - current_start
+    if new_stop is None:
+        new_stop = current_stop
+    assert new_stop <= current_stop
+    stop_diff_from_start = (new_stop - current_start).total_seconds()
+
+    # Crop each signal
+    signals = []
+    for i in range(len(edf.getSignalHeaders())):
+        # Question: should we use `_get_sample_frequency` instead?
```
raphaelvallat, line 805: Ping @skjerns what's the best way to get the sampling frequency?

skjerns: getSampleFrequency should be good for now

pyedflib/tests/test_highlevel.py
skjerns: no problem

pyedflib/tests/test_highlevel.py
```diff
@@ -292,32 +292,51 @@ def test_anonymize(self):
...
+        edf_file = os.path.join(data_dir, 'test_generator.edf')
+        outfile = os.path.join(data_dir, 'tmp_test_generator_cropped.edf')
+        new_start = datetime(2011, 4, 4, 12, 58, 0)
+        new_stop = datetime(2011, 4, 4, 13, 0, 0)
+        highlevel.crop_edf(
+            edf_file, new_file=outfile, new_start=new_start, new_stop=new_stop)
+        highlevel.crop_edf(
+            edf_file, new_file=outfile, new_start=None, new_stop=new_stop, verbose=False)
+        highlevel.crop_edf(
+            edf_file, new_file=outfile, new_start=new_start, new_stop=None)
```
skjerns, line 315: Can you add tests here that read the files and check for the timestamp of start/stop? :)

raphaelvallat:
Thanks for the review @skjerns. Sounds good about adding the possibility to crop based on samples or seconds. For simplicity, what about:

```python
highlevel.crop_edf(
"data/test_generator.edf", 
start=new_start, stop=new_stop, 
start_format="datetime", stop_format="datetime", 
verbose=True
)
```

where `start/stop_format` can be one of `"datetime"`, `"seconds"`, or `"samples"`?

skjerns:
Yes that sounds like a good solution!

pyedflib/tests/test_highlevel.py
raphaelvallat: Done

skjerns:
> Ready for re-review @skjerns !
> 
> Of note, I did not add support for "samples" since I think it could lead to errors when the signals have different sampling frequencies. The only supported options are "datetime" and "seconds". I have also updated the unit tests, which now compare that the signal values and headers are the same.

ah, yes that makes total sense. Thanks for thinking of this! I'll review it soon.

pyedflib/highlevel.py
```diff
@@ -766,6 +767,121 @@ def anonymize_edf(edf_file, new_file=None,
...
+    current_stop = current_start + timedelta(seconds=edf.getFileDuration())
+    current_duration = current_stop - current_start
+    if stop is None:
+        stop = current_stop
+    else:
+        if stop_format == "seconds":
+            stop = current_start + timedelta(seconds=stop)
+        else:
+            pass
+    assert stop <= current_stop
```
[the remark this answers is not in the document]
skjerns: do we need to check if `stop>current_start` as well?

pyedflib/highlevel.py
```diff
@@ -766,6 +767,121 @@ def anonymize_edf(edf_file, new_file=None,
...
+    # Define new start time
+    current_start = edf.getStartdatetime()
+    if start is None:
+        start = current_start
+    else:
+        if start_format == "seconds":
+            start = current_start + timedelta(seconds=start)
+        else:
+            pass
+    assert current_start <= start
```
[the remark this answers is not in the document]
skjerns: do we need to check if `start<current_stop` as well?

pyedflib/highlevel.py
```diff
@@ -766,6 +767,121 @@ def anonymize_edf(edf_file, new_file=None,
...
+    stop_diff_from_start = (stop - current_start).total_seconds()
+
+    # Crop each signal
+    signals = []
+    for i in range(len(edf.getSignalHeaders())):
+        sf = edf.getSampleFrequency(i)
+        # Convert from seconds to samples
+        start_idx = int(start_diff_from_start * sf)
+        stop_idx = int(stop_diff_from_start * sf)
+        # We use digital=True in reading and writing to avoid precision loss
```
skjerns, line 857:
```suggestion
        start_idx = int(np.round(start_diff_from_start * sf))
        stop_idx = int(np.round(stop_diff_from_start * sf))
        # We use digital=True in reading and writing to avoid precision loss
```

rounding first should be safer, the sample frequency is a shaky concept in EDF as it's calculated using recordlen smp_per_record, which can result in odd values of e.g. 199.9999 which should be `200` and not `199`.

pyedflib/highlevel.py
```diff
@@ -766,6 +767,121 @@ def anonymize_edf(edf_file, new_file=None,
...
+        new_file = file + "_cropped" + ext
+    write_edf(new_file, signals, signals_headers, header, digital=True)
+
+    # Safety check: are we able to load the new EDF file?
+    # Get new EDF start, stop and duration
+    edf = pyedflib.EdfReader(new_file)
+    start = edf.getStartdatetime()
+    stop = start + timedelta(seconds=edf.getFileDuration())
+    duration = stop - start
+    edf.close()
```
skjerns, line 876:
```suggestion
    with pyedflib.EdfReader(new_file) as edf:
        start = edf.getStartdatetime()
        stop = start + timedelta(seconds=edf.getFileDuration())
        duration = stop - start
        edf.close()
```

pyedflib/tests/test_highlevel.py
```diff
@@ -11,7 +11,30 @@
...
+    stop = start + timedelta(seconds=duration)
+
+    # Compare signal headers
+    assert signal_headers == orig_signal_headers
+
+    # Compare signal values
+    for i in range(signals.shape[0]):
+        sf_sig = signal_headers[i]["sample_frequency"]
+        idx_start = int((start - orig_start).seconds * sf_sig)
+        idx_stop = int((stop - orig_start).seconds * sf_sig)
```
skjerns, line 35:
```suggestion
        idx_start = int(np.round((start - orig_start).seconds * sf_sig))
        idx_stop = int(np.round((stop - orig_start).seconds * sf_sig))
```

raphaelvallat:
All suggestions merged. Thanks!

skjerns:
Thanks a lot for the contribution @raphaelvallat !

PR #200, opened 2023-05-08
remove mention of wavelet in doc

alistairewj:
Seems to be a typo in the index RST


skjerns:
okay, that is really weird. How did that end up there?

Generally the docs could need an overhaul and extension.

Thanks!

PR #203, opened 2023-06-02
Upgrade Python code to Python 3

DimitriPapadopoulos:
Apply most of the suggestions of [pyupgrade](https://github.com/asottile/pyupgrade).

The Python code is documented to  support Python >= 3.7 in [README.rst](https://github.com/holgern/pyedflib/blob/master/README.rst):
> pyEDFlib can be used with [Python](http://python.org/) >=3.7.


skjerns:
looking good, thanks!

could you merge the changes I made in #206 (now in master) and check if the test suit then runs?

skjerns:
thanks!

PR #204, opened 2023-06-02
3.7 <= Python <= 3.11

DimitriPapadopoulos:
- [README.rst](https://github.com/holgern/pyedflib/blob/master/README.rst) already states
  > pyEDFlib can be used with [Python](http://python.org/) >=3.7.
- Python 3.5 [reached EOL](https://devguide.python.org/versions/) in 2020.
- Python 3.6 [reached EOL](https://devguide.python.org/versions/) in 2021.
- Python 3.5 is not (easily) available in CI runners.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


skjerns:
lgtm!

PR #208, opened 2023-06-02
Consistent line terminators

DimitriPapadopoulos:
Change a single file from Windows **CR** **LF** to POSIX **LF**, for consistency.


skjerns:
tbh I did not even know this file exitst

PR #209, opened 2023-06-02
Apply isort

DimitriPapadopoulos:
Not that I necessarily agree with all [isort](https://pycqa.github.io/isort/) choices, but the result is consistent and readable imports


skjerns:
I think that's a good idea - would you be willing to update this for the current master branch?

DimitriPapadopoulos:
Done. This time I actually ran `ruff check --select I001` instead of `isort`.

PR #210, opened 2023-06-02
Update authors.py

DimitriPapadopoulos:
From version 1.10.1 of Scipy:
https://github.com/scipy/scipy/blob/c1ed5ec/tools/authors.py

I started up by trying to remove obsolete Python 2 stuff:
```
from scipy._lib.six import u, PY3
```
Ended up sync'ing this vendored file with upstream.

Do you really need it, by the way? The previous version wouldn't have worked with recent versions of SciPy any way, as they lack `scipy._lib.six`.

### DIFF
--- a/util/authors.py
+++ b/util/authors.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python
+# -*- encoding:utf-8 -*-
 """
-git-authors [OPTIONS] REV1..REV2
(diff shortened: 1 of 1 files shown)


skjerns:
tbh I don't even know how this file was used originally. I never touched it since starting to maintain this repo.

@holgern do you know anything?

PR #214, opened 2023-06-05
Update Unicode handling for Python 3

DimitriPapadopoulos:


pyedflib/edfreader.py
```diff
@@ -126,7 +126,7 @@ def _convert_string(self,s):
         if isinstance(s, bytes):
             return s.decode("latin")
         else:
-            return s.decode("utf-8", "strict")
+            return s.decode("utf_8", "strict")
```
skjerns, line 129: should this not be equivalent?

util/gh_lists.py
```diff
@@ -119,7 +119,6 @@ def __init__(self, filename):
             self.cache = {}
 
     def get(self, url):
-        url = unicode(url)
```
skjerns, line 122: tbh I don't even know what this file does.

skjerns:
Thanks! looks good :) could you squash the commits to a single commit with a meaningful message? that helps to keep the commit history a bit cleaner.

pyedflib/edfreader.py
DimitriPapadopoulos: Sure, but the reference name for the UTF-8 codec is `utf_8`:
https://docs.python.org/3/library/codecs.html#standard-encodings

The rest are aliases, `utf-8` is not even in the list of aliases:
> **CPython implementation detail:** Some common encodings can bypass the codecs lookup machinery to improve performance. These optimization opportunities are only recognized by CPython for a limited set of (case insensitive) aliases: utf-8, utf8, latin-1, latin1, iso-8859-1, iso8859-1, mbcs (Windows only), ascii, us-ascii, utf-16, utf16, utf-32, utf32, and the same using underscores instead of dashes. Using alternative aliases for these encodings may result in slower execution.

DimitriPapadopoulos:
Done.

PR #227, opened 2023-06-26
Update setup.py

wenh06:
Increment micro version by 1, so that push to PyPI action would succeed.

Ref. the [Build wheel action](https://github.com/holgern/pyedflib/actions/runs/5362811169/jobs/9730443796)

image


wenh06:
@holgern @skjerns

skjerns:
thanks! my bad, forgot that.

PR #229, opened 2023-06-27
introduces → introduced

DimitriPapadopoulos:


skjerns:
For a cleaner commit history it would be better to include such atomic changes either into other PRs or collect them in a larger PR. I think there are a lot of spelling mistakes to be found, but a general rewrite of the docs would be preferable (there's a bit of outdated stuff there and also more examples could be added) than fixing those minor mistakes.

DimitriPapadopoulos:
I understand, but I cannot rewrite the documentation soon, and I find this typo early in the README visible enough to keep me from reading smoothly.

skjerns:
Sure, that makes sense!

PR #231, opened 2023-06-28
Avoid deprecated sample_rate in our own code

### LINKED ISSUE #198: `'sample_rate'` is deprecated, but pyedflib still uses it internally
I am using `highlevel.write_edf()` in my tests, and even though I am passing `sample_frequency` instead of `sample_rate`, I still get a `DeprecationWarning`:

```
    def test_read_shhs(tmp_path):
        """Basic sanity checks for records read via read_shhs."""
        durations = [0.1, 0.2]  # hours
        valid_stages = {int(s) for s in SleepStage}
    
>       _create_dummy_shhs(data_dir=tmp_path, durations=durations)

sleepecg/test/test_sleep_readers.py:157: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sleepecg/test/test_sleep_readers.py:124: in _create_dummy_shhs
    _dummy_nsrr_edf(f"{edf_dir}/{record_id}.edf", hours, ecg_channel="ECG")
sleepecg/test/test_sleep_readers.py:29: in _dummy_nsrr_edf
    highlevel.write_edf(filename, ecg, signal_headers)
.direnv/python-3.10.10/lib/python3.10/site-packages/pyedflib/highlevel.py:486: in write_edf
    f.setSignalHeaders(signal_headers)
.direnv/python-3.10.10/lib/python3.10/site-packages/pyedflib/edfwriter.py:351: in setSignalHeaders
    self.update_header()
.direnv/python-3.10.10/lib/python3.10/site-packages/pyedflib/edfwriter.py:249: in update_header
    self._calculate_optimal_record_duration()
.direnv/python-3.10.10/lib/python3.10/site-packages/pyedflib/edfwriter.py:922: in _calculate_optimal_record_duration
    all_fs = [self._get_sample_frequency(i) for i,_ in enumerate(self.channels)]
.direnv/python-3.10.10/lib/python3.10/site-packages/pyedflib/edfwriter.py:922: in <listcomp>
    all_fs = [self._get_sample_frequency(i) for i,_ in enumerate(self.channels)]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyedflib.edfwriter.EdfWriter object at 0x7ff705cb0370>, channelIndex = 0

    def _get_sample_frequency(self, channelIndex):
        # Temporary conditional assignment while we deprecate 'sample_rate' as a channel attribute
        # in favor of 'sample_frequency', supporting the use of either to give
        # users time to switch to the new interface.
        if 'sample_rate' in self.channels[channelIndex]:
>           warnings.warn("`sample_rate` is deprecated and will be removed in a future release. \
                          Please use `sample_frequency` instead", DeprecationWarning)
E           DeprecationWarning: `sample_rate` is deprecated and will be removed in a future release.                           Please use `sample_frequency` instead
```

Currently, users cannot really prevent that warning even if they have already switched to the new `sample_frequency`.

DimitriPapadopoulos:
Fixes #198.

At least I think it should. Let's wait for feedback.

The idea is that we should be **able to read** files with `sample_rate` for backwards compatibility, but **not write** files with `sample_rate`. Of course, this means that older versions of pyedflib won't be able to read files produced with newer versions of pyedeflib. Hence, I guess, the CI failure:
```
            for shead1, shead2 in zip(signal_headers1, signal_headers2):
                # When only 'sample_rate' is present, we use its value to write
                # the file, ignoring 'sample_frequency', which means that when
                # we read it back only the 'sample_rate' value is present.
>               self.assertDictEqual({**shead1, 'sample_frequency': shead1['sample_rate']},
                                     shead2)
E               KeyError: 'sample_rate'
```

An alternative would be to output warnings when reading files, **only** when the values of `sample_rate` and  `sample_frequency` are different.

Thoughts?

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -171,19 +171,6 @@ class EdfWriter:
             edflib.FILETYPE_EDFPLUS
             edflib.FILETYPE_BDFPLUS
         n_channels is the number of channels without the annotation channel
-
-        channel_info should be a
-        list of dicts, one for each channel in the data. Each dict needs
-        these values:
-
-            'label' : channel label (string, <= 16 characters, must be unique)
-            'dimension' : physical dimension (e.g., mV) (string, <= 8 characters)
-            'sample_rate' : sample frequency in hertz (int). Deprecated: use 'sample_frequency' instead.
-            'sample_frequency' : number of samples per record (int)
-            'physical_max' : maximum physical value (float)
-            'physical_min' : minimum physical value (float)
-            'digital_max' : maximum digital value (int, -2**15 <= x < 2**15)
-            'digital_min' : minimum digital value (int, -2**15 <= x < 2**15)
         """
         self.path = file_name
         self.file_type = file_type
@@ -205,13 +192,15 @@ class EdfWriter:
         self.sample_buffer = []
         for i in np.arange(self.n_channels):
             if self.file_type == FILETYPE_BDFPLUS or self.file_type == FILETYPE_BDF:
-                self.channels.append({'label': f'ch{i}', 'dimension': 'mV', 'sample_rate': 100,
-                                      'sample_frequency': None, 'physical_max': 1.0, 'physical_min': -1.0,
+                self.channels.append({'label': f'ch{i}', 'dimension': 'mV',
+                                      'sample_frequency': 100,
+                                      'physical_max': 1.0, 'physical_min': -1.0,
                                       'digital_max': 8388607,'digital_min': -8388608,
                                       'prefilter': '', 'transducer': ''})
             elif self.file_type == FILETYPE_EDFPLUS or self.file_type == FILETYPE_EDF:
-                self.channels.append({'label': f'ch{i}', 'dimension': 'mV', 'sample_rate': 100,
-                                      'sample_frequency': None, 'physical_max': 1.0, 'physical_min': -1.0,
+                self.channels.append({'label': f'ch{i}', 'dimension': 'mV',
+                                      'sample_frequency': 100,
+                                      'physical_max': 1.0, 'physical_min': -1.0,
                                       'digital_max': 32767, 'digital_min': -32768,
                                       'prefilter': '', 'transducer': ''})
 
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -314,6 +314,19 @@ class EdfWriter:
             'digital_max' : maximum digital value (int, -2**15 <= x < 2**15)
             'digital_min' : minimum digital value (int, -2**15 <= x < 2**15)
         """
+        try:
+            sample_rate = channel_info.pop('sample_rate')
+        except KeyError:
+            pass
+        else:
+            if 'sample_frequency' in channel_info:
+                if sample_rate != channel_info['sample_frequency']:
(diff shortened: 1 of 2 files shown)


cbrnr:
I still get the same error when running the test. I hope I installed the branch correctly:

```
pip install -U git+https://github.com/DimitriPapadopoulos/pyedflib@make_signal_header#egg=pyedflib
```

DimitriPapadopoulos:
Not sure how to install a specific branch with `pip`. I installed myself from a repository clone with `python setup.py develop`.

@skjerns How do you think compatibility should be handled?
1. Should we **write** files with both `sample_frequency` and `sample_rate`, so that legacy software using older versions of pyedflib can still read new files?
2. If the answer to the above is yes, when reading files, shouldn't we emit a deprecation warning **only** when `sample_frequency` is missing or if the values of `sample_frequency` and `sample_rate` are different? Otherwise, we get a deprecation warning when reading our own files we have just written.

skjerns:
> 1. Should we **write** files with both `sample_frequency` and `sample_rate`, so that legacy software using older versions of pyedflib can still read new files?

I think there is a confusion - in all cases, `sample_rate` is written to the file, as is required by the EDF standard (there is no `sample_frequency` in EDF). The usage of `sample_frequency` is only used to avoid confusion, as the definition of `sample_rate` is not intuitive. Internally, `sample_frequency` is always converted to `sample_rate`.

> 2. If the answer to the above is yes, when reading files, shouldn't we emit a deprecation warning **only** when `sample_frequency` is missing or if the values of `sample_frequency` and `sample_rate` are different? Otherwise, we get a deprecation warning when reading our own files we have just written.

I think we should only give this warning if the user explicitly sets `sample_rate`. If subfunctions write sample_rate, that should be no problem.

That means:
* Raise `DeprecationWarning` only if the user explicitly tries to set it.
* Raise no warning if subfunctions use `sample_rate` without user interference.

DimitriPapadopoulos:
Indeed, an [EDF](https://www.edfplus.info/specs/edf.html) file contains:
>  8 ascii : duration of a data record, in seconds 
> 4 ascii : number of signals (ns) in data record 

> ns * 8 ascii : ns * nr of samples in each data record

Not sure I understand how it works out for several non-contiguous recordings. Any way, I will let you figure out an elegant solution, unless you want help/input on the subject.

skjerns:
superseeded by #268

PR #232, opened 2023-07-03
More documentation for "sample frequency"

### LINKED ISSUE #199: sampling rate is not accurate when data record duration is not 1 second
I opened an EDF file where the data record duration was set to 2 seconds.

In the signal headers, it said that the channels frequency was 512, which is incorrect as it is actually 256 samples per second. It seems like it returns the number of samples per data record and not per second. Is that intended? I did not find a way to read the data record duration using pyedflib so i don't see a way to get the real sampling rate per second in this situation.

If i open the file with mne, it returns 256 as the 'sfreq' property which is correct as an example.

DimitriPapadopoulos:
It turns out "sample frequency" is actually the number of samples per channel/signal in EDFlib:
https://www.teuniz.net/edflib/doc/edflib_8h.html#a11d03849cecc17a2599125a1a102b69e

We inherit this strange definition, and need to explain it.

Fixes #199.

### DIFF
--- a/pyedflib/edfreader.py
+++ b/pyedflib/edfreader.py
@@ -370,11 +370,10 @@ class EdfReader(CyEdfReader):
 
     def getBirthdate(self, string=True):
         """
-        Returns the birthdate as string object
+        Returns the birthdate as a string.
 
         Parameters
         ----------
-        None
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


skjerns:
I see. This is all a big mess.

I would still favour to express `sampling_frequency` in Hz (samples/second) and not in samples/block_size and therefore deviate from the definition in `edflib`. 
I don't think anybody thinks in terms of samples/block_size, as the block_size is irrelevant outside of the technicial file definitions of the EDF file.

skjerns:
superseeded by #268

Issue #233, opened 2023-07-18
Build fails with latest Cython (3.0.0)


hofaflo:
With yesterday's Cython release, building pyedflib seems to fail. Executing `pip install -e .` (Ubuntu 22.04.2, Python 3.10.9) produces the output shown below. Capping Cython to `"cython<3.0.0"` in `pyproject.toml` makes it work again, but I don't know if that's the best course of action :D

```
Obtaining file:///home/florian/coding/pyedflib
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [227 lines of output]
      /tmp/pip-build-env-i1rg6twi/overlay/lib/python3.10/site-packages/Cython/Compiler/Main.py:381: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /home/florian/coding/pyedflib/pyedflib/_extensions/_pyedflib.pyx
        tree = Parsing.p_module(s, pxd, full_module_name)
      warning: pyedflib/_extensions/edf.pxi:27:0: The 'DEF' statement is deprecated and will be removed in a future Cython version. Consider using global variables, constants, and in-place literals instead. See https://github.com/cython/cython/issues/4310
      [... lots of repetitions ...]
      warning: pyedflib/_extensions/edf.pxi:88:0: The 'DEF' statement is deprecated and will be removed in a future Cython version. Consider using global variables, constants, and in-place literals instead. See https://github.com/cython/cython/issues/4310
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      
      #from c_edf cimport *
      import locale
      import os
      import warnings
      cimport c_edf
              ^
      ------------------------------------------------------------
      
      pyedflib/_extensions/_pyedflib.pyx:24:8: 'c_edf.pxd' not found
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
          needed according the montage and filter settings, then display the data.
      
          """
      
      
          cdef c_edf.edf_hdr_struct hdr
               ^
      ------------------------------------------------------------
      
      pyedflib/_extensions/_pyedflib.pyx:136:9: 'edf_hdr_struct' is not a type identifier
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      
[...]

PR #234, opened 2023-07-19
hotfix Cython<3, else can't install

skjerns:
With newest Cython version 3.0, something with finding the modules is broken. Until  we have time to dive for a solution, I propose this hotfix to set Cython<3. As the Build is done in its own env, this should have no impact on the base env that we are installing to, ie. no side-effects

See #233

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 nose
-cython
+cython<3.0
 numpy
 pytest
 dateparser
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,6 @@ requires = [
     "setuptools",
     "wheel",
 	"oldest-supported-numpy",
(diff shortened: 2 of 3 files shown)


OlivierKessler01:
Please merge this asap, using --no-build-isolation doesn't make for a proper fix since the cython version would conflict with other libraries compiling needs.

skjerns:
However, install should still be possible using the wheels on pypi, or do you require the dev version?

OlivierKessler01:
> However, install should still be possible using the wheels on pypi, or do you require the dev version?
> 
> (I'm not sure if there is a way to install using pypi and having to use Cython?)

We are running 0.1.20 on musl based platforms, so no wheels available at the time, had I checked I would've seen that more recent versions of pyedflib provided wheels for us.

Cheers ! :)

PR #239, opened 2023-09-26
Remove wheel from pyproject.toml

cbrnr:
This dependency is automatically provided by `setuptools`(see https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a).


skjerns:
thanks!

PR #244, opened 2023-10-09
Add macOS ARM64 wheels

cbrnr:
I noticed that macOS ARM64 wheels are missing, so hopefully this PR will add them. Is there a way to test if this works on cibuildwheel?

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -56,6 +56,7 @@ jobs:
         env:
           # Disable  building PyPy wheels
           CIBW_SKIP: "pp* cp312*i686"
+          CIBW_ARCHS_MACOS: "x86_64 arm64"
           CIBW_PRERELEASE_PYTHONS: False
 
       - uses: actions/upload-artifact@v3


skjerns:
yes, but only on branches from the same repo. I added them here #245 , it seems to work

cbrnr:
Are the wheels available for download so that I can test installing them locally? I don't see the output...

cbrnr:
But I meant how do you know that the wheels are building OK when they are not being built? Or are they, and there are just no artifacts to download?

cbrnr:
Where do you see that they were successfully built?

skjerns:
> Where do you see that they were successfully built?

https://github.com/holgern/pyedflib/pull/244#issuecomment-1753449840

skjerns:
> Where do you see that they were successfully built?

https://github.com/holgern/pyedflib/actions/runs/6459073038/job/17534120890

cbrnr:
Thanks! But that happens only when you create a new tag, right? Would it not be useful to build all wheels for every PR? I mean, if you want to avoid a situation where wheels fail, then you would have to create another tag just to fix the wheels.

skjerns:
The wheel building can also be started by workflow dispatch, but then the wheels wont upload to pypi, that only works if the version is incremented :)

I tested the wheels build on a branch and once that was confirmed, I created the tag

cbrnr:
Ah, got it!

PR #266, opened 2024-12-11
Converted assert of phys_min and phys_max into warnings

TrianecT-Wouter:
Converted assert's into warnings. 

Reason:
in many cases, storing an edf file fails because of roundoff errors: e.g. `phys_min -3565.82 vs. signal_min 3565.8200000000000006` will fail in write_edf (but shouldn't)

### DIFF
--- a/pyedflib/highlevel.py
+++ b/pyedflib/highlevel.py
@@ -531,12 +531,12 @@ def write_edf(
             assert pmin != pmax, \
             f'physical_min {pmin} should be different from physical_max {pmax}'
         else: # only warning, as this will not lead to clipping
-            assert pmin<=sig.min(), \
-            'phys_min is {}, but signal_min is {} ' \
-            'for channel {}'.format(pmin, sig.min(), label)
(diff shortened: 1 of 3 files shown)


skjerns:
nice! It seems like it was originally intended to be a warning, so good fix.

TrianecT-Wouter:
Didn't see that before. Will check.

TrianecT-Wouter:
Good tips @skjerns. I attempted to implement the tests as well and found some new edge cases which should still throw an error.

Workflow is now as follows:
1. check if the difference is large (using accuracy based on the fact that edf uses 16 bits for the range from -physmin / +physmax)
2. Yes: perform assertion
3. No: throw warning

Also reflected in tests.

To enable building the python package, I had to remove to references to **distools** (which is deprecated; see #267)

This also fixes #267 in the sense that you can still build the pypi package using python 3.12.

skjerns:
thanks! looks good to me :)

Issue #242, opened 2023-10-05
Potential bug using setDatarecordDuration


withmywoessner:
Hello! I was experimenting with the `setDatarecordDuration()` in `EdfWriter`.
From the [documentation ](https://pyedflib.readthedocs.io/en/latest/ref/edfwriter.html#pyedflib.EdfWriter.setDatarecordDuration)it says:

> Sets the datarecord duration. The default value is 100000 which is 1 second. ATTENTION: the argument “duration” is expressed in units of 10 microSeconds!

But, I tried running the following code:
```
import pyedflib
f = pyedflib.EdfWriter(path,1)
f.setDatarecordDuration(100000)
f.close()
```
and got the following error:
<img width="773" alt="image" src="https://github.com/holgern/pyedflib/assets/20428109/d7879698-fa75-42e9-8e90-70e9d97b81b0">
Is this a bug or am I misunderstanding how the function works?
Thank you!

skjerns:
Mhh, I think this might be a bit more complicated, so unless you feel confident diving into the code base there's not much to do right now. But feel free to look into it!

It supports values from 0.01 to 60 afaik. Do you know what the `record_duration` parameter does? It simply sets the block size of the EDF file, so it's just a technical term for EDF-insides

skjerns:
Superseeded by #268 

My recommendation is to not use `setDatarecordDuration`, but if you need, you can use it with `seconds`, which is now also reflected in the docstring.

```
        Sets the datarecord duration. The default value is 1 second.
        The datarecord duration must be in the range 0.00001 to 60 seconds.
        Usually, the datarecord duration is calculated automatically to
        ensure that all sample frequencies are representable, nevertheless,
        you can overwrite the datarecord duration manually. This can, however,
        lead to unexpected side-effects in the sample frequency calculations.
```

PR #268, opened 2025-01-18
Deprecate sample_rate

### LINKED ISSUE #198: `'sample_rate'` is deprecated, but pyedflib still uses it internally

### LINKED ISSUE #247: sample_rate vs sample_frequency: how to continue?
There is confusions about the two terms and how they are implemented in `pyedflib` and also `edflib`. I'd like you'r opinion on how to continue. @DimitriPapadopoulos @cbrnr @holgern @raphaelvallat @LucaCerina @cfranklin11 

### EDF specs

In the [specs](https://www.edfplus.info/specs/edf.html) the following is defined:
*` 8 ascii : duration of a data record, in seconds` -> we call this internally `record_duration`
*`ns * 8 ascii : ns * nr of samples in each data record (per channel)` -> we call this internally `smp_per_record`

There is no `sample_frequency` or `sample_rate` within the EDF specs. However, these can be derived by dividing `smp_per_record` by `record_duration`.

### edflib

[edflib](https://www.teuniz.net/edflib/doc/edflib_8h.html#a11d03849cecc17a2599125a1a102b69e), introduces `samplefrequency` (equal to `smp_per_record/record_duration`, not always Hz) and `samplerate` (Hz), which is not actually implemented to set, just mentioned in the docstrings. There's an inconsistency as `Hz` is defined as `cycle per second`, but this isn't always reflected in `edflib`:

- > For example, if you want to use a samplerate of 0.5 Hz, set the samplefrequency to 5 Hz and the datarecord duration to 10 seconds, or set the samplefrequency to 1 Hz and the datarecord duration to 2 seconds."
  

### pyedflib

In `pyedflib`, `sample_rate` was initially like `edflib`'s `samplefrequency`. Later, `sample_frequency` was introduced to represent Hz, simplifying the user experience. The definition of `sample_frequency` in Hz is also how it is done in `[python-edf](https://github.com/cleemesser/python-edf). 

### Solutions

In principle I think nobody cares about the `record_duration` or how things are implemented internally. People want to set and retrieve a sampling rate in Hz and be done. You should not have to think about the `record_duration` unless you have good reason to. Imo the decision to be made is which of the two terms to keep.

**Options:**

A) Keep both terms, eventually removing `sample_rate`. This is user-friendly but inconsistent with `edflib`.

B) Revert to `sample_rate` only, aligning with `edflib` but disrupting recent changes. Treat `sample_rate` as Hz.

C) Align with `edflib`: use `sample_frequency` for `smp_per_record` and `sample_rate` for Hz, which could confuse users.

Personally I think A&B seem to be the best options, with A the least disrupting. If the unfortunate introduction of `sample_frequency` would not have been done, I would favour B, but that's too late now. What do you think?

skjerns:
As discussed in https://github.com/holgern/pyedflib/issues/247 , I deprecated `sample_rate` and replaced all of its occurences with `sample_frequency`.

Frequency is usually measured in Hz, which is defined per second and is most consistent with other libraries (e.g. `mne`).

**Summary**
- There are errors raised when users try to use `sample_rate` - this is to ensure everyone finds out about the change without googeling
- I added a new routine to calculate `record_duration` - it should find an optimal value if one exists and let users not worry about the `sample_rate` as it is defined in `edflib` but to simply sey `sample_frequency` and be done
- It's still possible to overwrite the value of `record_duration` if ppl actually need to for obscure reasons

Additionally, needed to deprecate 3.7 for the test suite as it's not available for ubuntu24

fixes #198
closes #247

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-2019, windows-latest, macos-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,13 +9,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-2019, windows-latest, macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
-        exclude:  # Python < v3.8 does not support Apple Silicon ARM64.
-          - python-version: "3.7"
-            os: macos-latest
-        include:  # So run those legacy versions on Intel CPUs.
-          - python-version: "3.7"
-            os: macos-13
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+
     steps:
     - uses: actions/checkout@v3
     - name: Setup Python
--- a/pyedflib/_extensions/_pyedflib.pyx
+++ b/pyedflib/_extensions/_pyedflib.pyx
@@ -555,12 +555,11 @@ def set_equipment(handle, equipment):
 def set_samples_per_record(handle, edfsignal, smp_per_record ):
     """
     int set_samples_per_record(int handle, int edfsignal, int smp_per_record )
-
-    sets how many samples are in each record for this signal.
-    this is not the sampling frequency (Hz), unless record_duration=1
-    The sample frequency is calculated by smp_per_record/record_duration.
-    The function call to the C library is therefore slightly mislabeled
     """
+    # sets how many samples are in each record for this signal.
+    # this is not the sampling frequency (Hz), unless record_duration=1
+    # The sample frequency is calculated by smp_per_record/record_duration.
+    # The function call to the C library is therefore slightly mislabeled
     return c_edf.edf_set_samplefrequency(handle, edfsignal, smp_per_record)
 
 def set_admincode(handle, admincode):
--- a/pyedflib/_extensions/_pyedflib.pyx
+++ b/pyedflib/_extensions/_pyedflib.pyx
@@ -123,7 +123,9 @@ cdef class CyEdfReader:
     Note that edflib.c is encapsulated so there is no direct access to the file
     from here unless I add a raw interface or something
 
-    EDF/BDF+ files are arranged into N signals sampled at rate Fs. The data is actually stored in chunks called    "datarecords" which have a file specific size.
+    EDF/BDF+ files are arranged into N signals sampled at rate Fs.
+    The data is actually stored in chunks called  "datarecords" which have a
+    file specific size.
 
     A typical way to use this to read an EEG file would be to choose a certain
     number of seconds per page to display. Then figureout how many data records
@@ -554,9 +556,10 @@ def set_samples_per_record(handle, edfsignal, smp_per_record ):
     """
     int set_samples_per_record(int handle, int edfsignal, int smp_per_record )
 
-    sets how many samples are in the record for this signal.
-    this is not the sampling frequency (Hz), (which is calculated by
-    by smp_per_record/record_duration).
+    sets how many samples are in each record for this signal.
+    this is not the sampling frequency (Hz), unless record_duration=1
+    The sample frequency is calculated by smp_per_record/record_duration.
+    The function call to the C library is therefore slightly mislabeled
     """
     return c_edf.edf_set_samplefrequency(handle, edfsignal, smp_per_record)
 
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -381,7 +381,7 @@ class EdfWriter:
         for ch in self.channels:
             # raise exception, can be removed in later release
             if 'sample_rate' in ch:
-                raise ValueError('Use of `sample_rate` is deprecated, use `sample_frequency` instead')
+                raise FutureWarning('Use of `sample_rate` is deprecated, use `sample_frequency` instead')
 
         sample_freqs = [ch['sample_frequency'] for ch in self.channels]
         if not self._enforce_record_duration and not any([f is None for f in sample_freqs]):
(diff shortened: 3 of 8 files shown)


skjerns:
@cbrnr @cfranklin11 @LucaCerina @raphaelvallat @DimitriPapadopoulos  if any of you has time to have a look, that would be great :)

raphaelvallat (approved):
LGTM 👍

LucaCerina:
Good also for me

pyedflib/_extensions/_pyedflib.pyx
```diff
@@ -554,9 +556,10 @@ def set_samples_per_record(handle, edfsignal, smp_per_record ):
...
     int set_samples_per_record(int handle, int edfsignal, int smp_per_record )
 
-    sets how many samples are in the record for this signal.
-    this is not the sampling frequency (Hz), (which is calculated by
-    by smp_per_record/record_duration).
+    sets how many samples are in each record for this signal.
+    this is not the sampling frequency (Hz), unless record_duration=1
+    The sample frequency is calculated by smp_per_record/record_duration.
+    The function call to the C library is therefore slightly mislabeled
     """
```
cbrnr, line 563: If this is supposed to be a comment, I suggest using an actual comment instead of a string.

pyedflib/edfwriter.py
```diff
@@ -292,7 +377,17 @@ def update_header(self) -> None:
         # the channels, we need to find a common denominator so that all sample
         # frequencies can be represented accurately.
         # this can be overwritten by explicitly calling setDatarecordDuration
-        self._calculate_optimal_record_duration()
+
+        for ch in self.channels:
+            # raise exception, can be removed in later release
+            if 'sample_rate' in ch:
+                raise ValueError('Use of `sample_rate` is deprecated, use `sample_frequency` instead')
```
cbrnr, line 384: Are you sure you don't want to raise a `FutureWarning` instead?

pyedflib/edfwriter.py
```diff
@@ -526,10 +620,11 @@ def setDatarecordDuration(self, record_duration: Union[float, int]) -> None:
         This function is NOT REQUIRED but can be called after opening a file
         in writemode and before the first sample write action. This function
         can be used when you want to use a samplefrequency which is not an
-        integer. For example, if you want to use a samplerate of 0.5 Hz, set
+        integer. For example, if you want to use a samplefreq of 0.5 Hz, set
```
cbrnr, line 623: I'd spell it out as "sample frequency" to be consistent.

.github/workflows/tests.yml
```diff
@@ -9,13 +9,8 @@ jobs:
...
       matrix:
         os: [ubuntu-latest, windows-2019, windows-latest, macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
-        exclude:  # Python < v3.8 does not support Apple Silicon ARM64.
-          - python-version: "3.7"
-            os: macos-latest
-        include:  # So run those legacy versions on Intel CPUs.
-          - python-version: "3.7"
-            os: macos-13
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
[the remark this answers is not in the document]
skjerns: sure!

pyedflib/edfwriter.py
skjerns: right! that makes sense :)

skjerns:
Found something strange: Even though the `record_duration` can be 8 ASCII chars according to the definition, it truncates after 7 in our library. 

`AssertionError: read 0.12345 != expected 0.12345678`

In the raw header there is an empty space

image

this issue might be related https://github.com/holgern/pyedflib/issues/242. I suggest fixing this in a future PR.

skjerns:
Given that the the requested changes were minor, I'll merge this one :) thanks for reviewing!

PR #269, opened 2025-01-18
Start applying ruff rules

DimitriPapadopoulos:
@skjerns Ready for review.


pyedflib/tests/test_edfwriter.py
```diff
@@ -442,16 +441,14 @@ def test_SampleWriting(self):
...
-        data_list.append(data2)
+        data_list = [data1, data2]
         f.writeSamples(data_list)
         f.close()
 
         f = pyedflib.EdfReader(self.bdfplus_data_file)
         data1_read = f.readSignal(0)
         data2_read = f.readSignal(1)
-        f._close
+        f._close()
```
skjerns, line 451: uh strange that this never gave any access/locking problems

pyedflib/tests/test_edfwriter.py
```diff
@@ -924,16 +914,17 @@ def test_sample_rate_deprecation(self):
         digMax = 32767
         digMin = -digMax
 
-        base_signal_header = lambda idx: {
```
skjerns, line 927: wait, why was this a lambda expression, this is a mistake

DimitriPapadopoulos:
Two options:
* use `pytest-ruff` within the existing test GitHub Action (that's what I've chosen in 88ec4c120defb00b5e1039e27aad5c025088b3d1
* create a new ruff job

skjerns:
> Two options:
> 
> * use `pytest-ruff` within the existing test GitHub Action (that's what I've chosen in [88ec4c1](https://github.com/holgern/pyedflib/commit/88ec4c120defb00b5e1039e27aad5c025088b3d1))
> * create a new ruff job

I think it would be nicer to have one dedicated job, that way it is easier to see if errors are caused by cosmetic changes or functional tests failing. 

Also using the current `matrix` layout the ruff test will run redundantly several times needlessly.

DimitriPapadopoulos:
Updated to run ruff directly from a GitHub Action.

PR #270, opened 2025-01-18
Remove Python 2.7 leftovers

### LINKED ISSUE #265: Check if du() and u() are still necessary
During code review we foudn this older function `du()` and `u()` for unicode encoding and decoding. I suspect they are of no use anylonger after deprecating python 2.7. This should be checked

https://github.com/holgern/pyedflib/blob/4c6b95b534279f4d3ba41c585f8fbad1c2a0cac1/pyedflib/edfwriter.py#L144

https://github.com/holgern/pyedflib/blob/4c6b95b534279f4d3ba41c585f8fbad1c2a0cac1/pyedflib/edfwriter.py#L140

DimitriPapadopoulos:
Fixes #265.

### DIFF
--- a/pyedflib/highlevel.py
+++ b/pyedflib/highlevel.py
@@ -753,7 +753,7 @@ def drop_channels(
             if isinstance(ch,str):
                 ch_idx = ch_names.index(ch.lower())
                 to_keep[i] = ch_idx
-        load_channels = to_keep.copy()
+        load_channels = list(to_keep)
     elif to_drop is not None:
         for i,ch in enumerate(to_drop):
             if isinstance(ch,str):
--- a/pyedflib/highlevel.py
+++ b/pyedflib/highlevel.py
@@ -753,7 +753,7 @@ def drop_channels(
             if isinstance(ch,str):
                 ch_idx = ch_names.index(ch.lower())
                 to_keep[i] = ch_idx
-        load_channels = list(to_keep) # copy list compatible with py2.7
+        load_channels = to_keep.copy()
     elif to_drop is not None:
         for i,ch in enumerate(to_drop):
             if isinstance(ch,str):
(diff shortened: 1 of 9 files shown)


DimitriPapadopoulos:
CI errors will be fixed by #266 / ab3c58c0b820ab2b8c73ac2ca5168538d91d6feb or #269 / 78ff00ed5f5aaba05dff00c027b871b6ecdcd6e4.

skjerns:
> CI errors will be fixed by #266 / [ab3c58c](https://github.com/holgern/pyedflib/commit/ab3c58c0b820ab2b8c73ac2ca5168538d91d6feb) or #269 / [78ff00e](https://github.com/holgern/pyedflib/commit/78ff00ed5f5aaba05dff00c027b871b6ecdcd6e4).

sure! let me know when I can review

DimitriPapadopoulos:
#265 should be fixed in its own PR. The current code needs to be cleaned up and it should be clarified whether header fields should be stored as `str` or `byte`. It's currently a messed up mix of both as far as I can understand.

skjerns:
Looks good, I made some small adjustements

How about https://github.com/holgern/pyedflib/issues/265 ?

pyedflib/highlevel.py
```diff
@@ -753,7 +753,7 @@ def drop_channels(
             if isinstance(ch,str):
                 ch_idx = ch_names.index(ch.lower())
                 to_keep[i] = ch_idx
-        load_channels = to_keep.copy()
+        load_channels = list(to_keep)
```
DimitriPapadopoulos, line 756: `to_keep.copy()` is more readable, why revert?

DimitriPapadopoulos: It's supposed to be a `list`. Why would it be a Numpy array?

PR #285, opened 2026-07-02
Add BufferedEdfWriter for arbitrary-size (streaming) sample blocks

### LINKED ISSUE #284: writeSamples() silently pads sub-record blocks, inflating file duration during incremental/streaming writes
### Summary

When `EdfWriter.writeSamples()` is called incrementally with blocks **shorter than one data record** (fewer than `fs` samples per signal, with the default 1 s record), each call commits a **full data record** whose trailing part is padded with a constant value. In a real-time / streaming acquisition loop that writes each device block as it arrives, this silently produces a file that is much longer than the real recording, with most samples being padding. No exception or warning is raised, and the result is a structurally valid EDF+ that opens normally in any reader.

This is technically consistent with the documented per-record semantics, but it is an easy and costly trap when the acquisition loop and the file writer are developed independently. It would be more intuitive for the writer to **accept input blocks of arbitrary size** and buffer internally until a full record is available.

### Minimal reproducible example

```python
import numpy as np
import pyedflib

fs = 1000                      # 1 kHz
sig = np.random.randn(10_000)  # 10 s of data

w = pyedflib.EdfWriter("naive.edf", 1)
w.setSignalHeader(0, {
    "label": "EMG", "dimension": "mV", "sample_frequency": fs,
    "physical_min": -5, "physical_max": 5,
    "digital_min": -32768, "digital_max": 32767,
    "transducer": "", "prefilter": "",
})
# Stream in 100-sample blocks (100 ms), as a real-time device would
for i in range(0, len(sig), 100):
    w.writeSamples([sig[i:i+100]])
w.close()

r = pyedflib.EdfReader("naive.edf")
print("file_duration:", r.file_duration, "s")   # -> 100.0 s for 10 s of data
r.close()
```

**Expected:** a ~10 s file.
**Actual:** a ~100 s file; only ~10% of samples carry real data, the rest is padding. Amplitude and spectral estimates read back from the file are badly distorted.

### Proposed fix

Two options, from least to most invasive:

1. **Documentation:** add a note in the docstring of `writeSamples()` that each call commits complete records, and that passing fewer than `fs` samples per signal pads the record; recommend a buffer-then-flush pattern for streaming.
2. **Internal buffering (preferred):** accept blocks of arbitrary length, accumulate pending samples internally, commit only complete records, and flush the remainder on `close()`, padding the final record only.

I would be happy to open a PR adding an opt-in `BufferedEdfWriter` wrapper (accepts arbitrary-size blocks, commits only complete records, pads only the final record with the last value) with tests, if that direction is welcome.

### Version

Observed with `pyedflib 0.1.42`.

aagisto-maker:
## Summary

Real-time acquisition loops usually read the device in small blocks (for example 10–100 samples). Passing such sub-record blocks directly to `EdfWriter.writeSamples` makes each call commit a full data record padded to length, which silently inflates the file duration and distorts the stored signal (see issue #284). This PR adds an **opt-in** `BufferedEdfWriter` wrapper that accepts blocks of arbitrary size, commits a record only once a full record is available, and pads just the final record (with the last acquired value) on `close()`. It does **not** change `EdfWriter`'s existing behaviour.

## What is included

- `pyedflib/buffered.py` — the `BufferedEdfWriter` class.
- `pyedflib/tests/test_buffered.py` — tests: buffering prevents the file-duration inflation, documents the naive sub-record inflation, and covers the multichannel case.

## Example

```python
import pyedflib
from pyedflib.buffered import BufferedEdfWriter

writer = pyedflib.EdfWriter("out.edf", 1)
writer.setSignalHeader(0, header)
writer.setDatarecordDuration(1.0)

bw = BufferedEdfWriter(writer)
while acquiring:
    bw.write_samples([device.read(100)])   # any block size is fine
bw.close()
```

## Notes

- Fully backward compatible: a new, opt-in class; no change to the existing API.
- If preferred, it could be exported from the top-level package so that `from pyedflib import BufferedEdfWriter` works; happy to add that.
- Closes #284.

### DIFF
--- a/pyedflib/buffered.py
+++ b/pyedflib/buffered.py
@@ -1,0 +1,82 @@
+"""Buffered EDF/BDF writing for real-time (streaming) acquisition.
+
+Real-time acquisition loops usually read the device in small blocks (for example
+10-100 samples). Passing such sub-record blocks directly to
+:meth:`pyedflib.EdfWriter.writeSamples` makes each call commit a full data record
+padded to length, which silently inflates the file duration and distorts the
+stored signal. :class:`BufferedEdfWriter` wraps an ``EdfWriter`` and accepts
+blocks of arbitrary size, committing a record only once a full record is
+available and padding just the final record on :meth:`close`.
+"""
+
+import numpy as np
+
+__all__ = ["BufferedEdfWriter"]
+
+
+class BufferedEdfWriter:
+    """Accept arbitrary-size sample blocks and commit only complete data records.
+
+    Parameters
+    ----------
+    edf_writer : pyedflib.EdfWriter
+        A writer whose signal headers and data-record duration are already set.
+
+    Examples
+    --------
+    >>> writer = EdfWriter("out.edf", 1)
+    >>> writer.setSignalHeader(0, header)
+    >>> writer.setDatarecordDuration(1.0)
+    >>> bw = BufferedEdfWriter(writer)
+    >>> while acquiring:
+    ...     bw.write_samples([device.read(100)])   # any block size is fine
+    >>> bw.close()
+    """
+
+    def __init__(self, edf_writer):
+        self._w = edf_writer
+        self._n = len(edf_writer.channels)
+        self._rec = [int(edf_writer.get_smp_per_record(i)) for i in range(self._n)]
+        self._buf = [np.array([], dtype=np.float64) for _ in range(self._n)]
+
+    def write_samples(self, data_list):
+        """Append one block per signal; flush every complete data record."""
+        if len(data_list) != self._n:
+            raise ValueError(
+                "expected {} signals, got {}".format(self._n, len(data_list))
+            )
+        for i in range(self._n):
+            self._buf[i] = np.concatenate(
+                [self._buf[i], np.asarray(data_list[i], dtype=np.float64)]
+            )
+        while all(len(self._buf[i]) >= self._rec[i] for i in range(self._n)):
+            record = [
+                np.ascontiguousarray(self._buf[i][: self._rec[i]])
+                for i in range(self._n)
+            ]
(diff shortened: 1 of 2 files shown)


skjerns:
Thanks for the addition!

I like the functionality, however, I think it would be better to not introduce another module with different caller signatures for this purpose. It should be possible to integrate this into the current `EdfWriter` code by adding a `buffered=False` as a default parameter and skip the steps after

https://github.com/holgern/pyedflib/blob/2626a9b23545a97392931c0ca57a6d2f59dd61ef/pyedflib/edfwriter.py#L977

Then, on `close()` we write the final samples. 

Additionally, change `pad_with_last` to `pad_with` that accepts either an `int/float` or `'last'` with 0 being the default (here we also need to pay attention to physical or digital 0 I realize). 

I've fired up a plan by Claude and make some changes to your PR:

```raw
1. __init__(..., buffered=False) — store the flag and per-channel leftover buffers. Amusingly,
self.sample_buffer already exists at edfwriter.py:324 and is completely dead code (also only populated in
the EDF branch due to an indentation slip at line 337) — it was apparently intended for exactly this and
could be repurposed.
2. In writeSamples, if buffered: prepend the stored leftovers (np.concatenate([buf[i], data_list[i]]))
before the record-size bookkeeping, run the while loop unchanged, then instead of the tail block, stash
data_list[i][ind[i]:] back into the buffer and return.
3. Factor the existing tail block (977–989) into a small _flush_last_record() helper; non-buffered
writeSamples calls it as today, and close() calls it when buffered and buffers are non-empty. Since
__exit__ and __del__ already delegate to close(), context-manager usage flushes automatically
```

skjerns:
can you check out https://github.com/holgern/pyedflib/pull/286 and see it that fits your purpose?

PR #286, opened 2026-07-18
Add buffered=True option to EdfWriter for streaming sub-record blocks

### LINKED ISSUE #284: writeSamples() silently pads sub-record blocks, inflating file duration during incremental/streaming writes

skjerns:
Reworks #285 (thanks @aagisto-maker - commits kept) so the buffering lives in `EdfWriter` itself instead of a separate wrapper class.

`EdfWriter(..., buffered=True)` makes `writeSamples()` accept blocks of arbitrary size: samples are buffered per channel and only complete data records are committed; the incomplete remainder is written zero-padded on `close()`. Default behaviour (`buffered=False`) is unchanged. This fixes the file-duration inflation from #284 when streaming small acquisition blocks.

Notes:
- Repurposes the previously unused `sample_buffer` attribute (and fixes it only being initialised for EDF file types).
- Mixing `digital=True/False` across buffered calls raises `TypeError`.
- Unlike #285, the final record is zero-padded for consistency with unbuffered `writeSamples`; a configurable `pad_with` (e.g. `'last'`) is planned as a follow-up PR.
- Tests merged into `test_edfwriter.py`.

Closes #284, closes #285.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skjerns:
@aagisto-maker could you checkout this branch and test if it suits your purposes? 

I'll add the `pad_with` in another PR then.

aagisto-maker:
Thanks for reworking this, Simon — and for keeping the commits cleaner to have it inside `EdfWriter` than the wrapper I proposed. I'll check out the branch and run it through my acquisition case, then report back early next week.

aagisto-maker:
Ran it through my acquisition case over the weekend — real-time EMG at 1 kHz, 3 channels, written in 100-sample (100 ms) device blocks, a 30 s recording:

- `buffered=False` (the old per-block `writeSamples`): header duration   inflates to 300 s (10×), 300 000 samples/channel — the #284 bug.
- `buffered=True` (this PR): header duration back to 30 s, 30 000   samples/channel. The recovered signal is bit-for-bit identical to a reference file written in full records, with max error vs. the original within one EDF quantization step. The 6 buffered tests pass locally too.

So it fully suits my purpose — happy for this to land.

One thought for the planned `pad_with` follow-up: `pad_with='last'` would be the more useful default for biopotential streams. Zero-padding the final partial record injects a step to/from 0 at the very end, which shows up as a spectral edge artifact in the envelope/PSD of the last segment. Padding with the last acquired value avoids that discontinuity.
Happy to help test that PR when it's up.

skjerns:
> One thought for the planned `pad_with` follow-up: `pad_with='last'` would be the more useful default for biopotential streams. Zero-padding the final partial record injects a step to/from 0 at the very end, which shows up as a spectral edge artifact in the envelope/PSD of the last segment. Padding with the last acquired value avoids that discontinuity. Happy to help test that PR when it's up.

I understand, but this would break backward-compatibility. Therefore, I'd like to keep the current default of writing zeros, even if it's suboptimal, else we'd have to send a warning for each new write and use a deprecation cycle.

aagisto-maker:
Makes sense — not worth a deprecation cycle over it. Keeping zeros as the default is fine; an opt-in `pad_with='last'` is all I need on my side. I'll set it explicitly for the EMG streams so the final record doesn't introduce an edge artifact. Happy to test the follow-up PR whenever it's up.

skjerns:
nice! feel free to review #300

aagisto-maker:
Checked #300 against my streaming case — buffered EMG at 1 kHz, 3 channels, 100-sample blocks, with a non-integer number of records so the final one is partial:

- Default (`pad_with=0`): unchanged — the final partial record is zero-padded, so it's backward-compatible.
- `pad_with='last'`: the trailing pad now takes each channel's last acquired value. With a signal sitting well off zero at the end,
  zero-padding leaves a ~1.5 mV step into the last record, while `pad_with='last'` removes it (step ≈ 0). Real samples are untouched in both modes, and the out-of-range warning fires as expected.

This is exactly what I needed for the biopotential case — I'll pass `pad_with='last'` for the EMG streams. Thanks for turning it around so quickly. LGTM.

PR #293, opened 2026-07-19
Drop support for Python 3.8

### LINKED ISSUE #292: Drop support for NumPy 1?
NumPy 1 and 2.0 have reached EOL and supported versions of NumPy require Python 3.12:
https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

Time to at least drop support for NumPy 1 and Python 3.8?

DimitriPapadopoulos:
Python 3.8 requires an obsolete version of pip that doesn't support PEP 639 and 735.

Fixes #292.


pyedflib/edfwriter.py
```diff
@@ -335,8 +335,8 @@ def __init__(self, file_name: str, n_channels: int, file_type: int = FILETYPE_ED
                 raise ValueError("pad_with must be a number or 'last', "
                                  f"got {pad_with!r}")
         elif isinstance(pad_with, bool) or not isinstance(pad_with, numbers.Number):
-            raise ValueError("pad_with must be a number or 'last', "
-                             f"got {pad_with!r}")
+            raise TypeError("pad_with must be a number or 'last', "
```
DimitriPapadopoulos, line 358: This slightly break backwards compatibility:
`ValueError` → `TypeError`

Is such a change acceptable?

skjerns: The change wasn't released anyway yet, so should be okay!

skjerns (approved):
looks good, thanks!

PR #295, opened 2026-07-19
No need to ignore these ruff rules

DimitriPapadopoulos:
* `F841`
* `RUF100`


skjerns:
No idea why these were ignored before anyway, thanks!

PR #296, opened 2026-07-19
Apply ruff/pyupgrade rules UP031 and UP032

DimitriPapadopoulos:


skjerns:
thanks!

I think a lot of these files are actually remnants from a previous period. We should clean up the repo at some point. Eg `demo` is not referenced anywhere and wildely out of date w.r.t to available functions. `util` also has quite a few unused scripts

DimitriPapadopoulos:
Can we start with removing demo?

skjerns:
Sure! However, sometimes examples are actually nice and useful. What do you think about renaming it to `/examples`  (as a more canonical name)  and add some more up-to-date examples of how to use the library?

PR #298, opened 2026-07-19
build: migrate from setup.py/setup.cfg to pyproject.toml

### LINKED ISSUE #225: setup.py → pyproject.toml
Move to `pyproject.toml` and make `setup.py` a thin wrapper

[Configuring setuptools using `pyproject.toml` files](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)

DimitriPapadopoulos:
Replace the imperative `setup.py` with a PEP 517 custom build backend (`_custom_build.py`) that wraps `setuptools.build_meta`. All project metadata is now declared in `pyproject.toml` under PEP 621 `[project]`.

The `_custom_build.py` hook handles the parts that cannot be expressed declaratively:
- static C library (c_edf) linked into the Cython extension
- lazy numpy include-dir resolution
- Windows UTF-8 patch (`edflib_utf8.c` generation)
- `version.py` regeneration at build time

Fixes #225.


pyproject.toml
```diff
@@ -1,10 +1,77 @@
...
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: C",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
skjerns, line 50:
```suggestion
    "Programming Language :: Python :: 3.13",
    "Programming Language :: Python :: 3.14",
```

pyproject.toml
```diff
@@ -1,10 +1,77 @@
...
+# ---------------------------------------------------------------------------
+
+[project]
+name = "pyEDFlib"
+version = "0.1.33"
+description = "library to read/write EDF+/BDF+ files"
+readme = "README.rst"
+license = { text = "BSD" }
+authors = [
+    { name = "Holger Nahrstaedt", email = "nahrstaedt@gmail.com" },
```
skjerns, line 29:
```suggestion
    { name = "Holger Nahrstaedt"},
    { name = "Simon Kern"},
    { name = "Dimitri Papadopoulos"},

```

pyproject.toml
```diff
@@ -1,10 +1,77 @@
...
+build-backend = "_custom_build"
+backend-path = ["."]
+
+# ---------------------------------------------------------------------------
+# Project metadata  (PEP 621)
+# ---------------------------------------------------------------------------
+
+[project]
+name = "pyEDFlib"
+version = "0.1.33"
```
skjerns:
```suggestion
version = "0.1.42"
```

Also any way to have one central version number somewhere and not in several sites?

_custom_build.py
```diff
@@ -0,0 +1,337 @@
...
+import sysconfig
+from functools import partial
+
+# ---------------------------------------------------------------------------
+# Version constants — single source of truth
+# ---------------------------------------------------------------------------
+
+MAJOR = 0
+MINOR = 1
+MICRO = 33
```
skjerns:
```suggestion
MICRO = 42
```

skjerns: this is the current `MICRO`, not sure why it still had `33`?

skjerns:
Thanks! This is long overdue. Can you solve the merge conflicts?

Two things: 

1. There should be one source of truth for the `version`. Either the `pyproject.toml` and retrieve in the `_custom_build.py`.
2. I think the force-push overwrote some of the RUF ignore changes from #296 ? Check that when merging

some minor comments from Claude:

```
1. Stale build.py naming. The file is _custom_build.py, but the docstring's own example
(_custom_build.py:12) says build-backend = "build" (wrong — would fail; actual is "_custom_build"), and
pyproject.toml:2-3 comments plus the generated version.py header (_custom_build.py:92) all say "build.py".
Confusing and, in the docstring's case, incorrect.

2. finalize_options monkey-patch stacks. _prepare_build() reassigns
setuptools.dist.Distribution.finalize_options on every hook call, capturing the already-patched function as
_orig_finalize. Since pip calls prepare_metadata_for_build_wheel then build_wheel, the wrappers nest. The
_pyedflib_injected guard prevents double-injection so it's not a bug today, but it globally mutates
setuptools and is fragile. Capture _orig_finalize once at import, or use a proper build_meta subclass.

3. Untested path: sdist → wheel. _custom_build.py is in MANIFEST.in (good — required for building from an
sdist), but since the backend now lives in a top-level file, CI should explicitly build the sdist and then
build a wheel from it. Worth adding before this lands, since wheels.yml publishes both
```

DimitriPapadopoulos:
Note that reading the version from `pyproject.toml` makes `tomli` a **build** depdendency for Python < 3.11. I'm working on this.

DimitriPapadopoulos:
Reading the version from `pyproject.toml` without `tomli` resolves the extra build dependency issue.

skjerns (commented):
I only see now that my comments have not been posted and were marked as pending, sorry!

pyproject.toml
DimitriPapadopoulos: This one will be fixed by rebasing.

pyproject.toml
DimitriPapadopoulos: I would rather do that in a different PR. Here I am merely translating as is from `setup.py` and applying new PEPs.

_custom_build.py
DimitriPapadopoulos: Ooops... Fixed while rebasing.

pyproject.toml
DimitriPapadopoulos: There is. I had asked an IA to do that, but somehow the commit got last while rebasing — probably because I have two directory checkouts in two different locations.

I'll just ask again.

DimitriPapadopoulos: Done, we now have a single source of truth.

doc/source/dev/how_to_release.rst
```diff
@@ -12,8 +12,10 @@ uploads them to PyPI using the ``PYPI_API_TOKEN`` repository secret.
 Bump the version
 ----------------
 
-Set the version in ``setup.py`` by editing ``MAJOR``, ``MINOR`` and ``MICRO``,
-and set ``ISRELEASED = True``. Commit the change.
+Set the version in ``pyproject.toml`` under the ``[project]`` section by
+editing the ``version`` field. The ``ISRELEASED`` status is determined
+automatically (``True`` if the version does not contain ``.dev``).
+Commit the change.
```
DimitriPapadopoulos: Is that acceptable? I don't know how else to handle `ISRELEASED` from `pyproject.toml`.

DimitriPapadopoulos:
Could you rerun Claude on this PR? I mistakenly erased a few changes while working from two different repositories and had to reapply the changes from scratch.

Also, the dev doc has been modified to account for the changes.

doc/source/dev/how_to_release.rst
DimitriPapadopoulos: Why not get the version directly from the release tag, using [setuptools-scm](https://pypi.org/project/setuptools-scm/)?

skjerns: yes, that would be great! would make things much easier.

would just need to check that this is compatible with the current pypi-release workflow, but I think it should

skjerns:
comments from claude (haven't looked at it myself yet, sorry!):

### Defects

**1. `requires-python` is not correct.**

`pyproject.toml` has `requires-python = ">=3.8"`. But commit d895d14 removed the support for Python 3.8, and the classifiers start at Python 3.9. Change the value to `>=3.9`.

**2. The version string of a development release is not correct.**

The code sets `ISRELEASED` to false when the version contains `.dev`. Then the code adds `.dev0+<hash>` a second time:

```
version = "0.1.43.dev0"   ->   full_version = "0.1.43.dev0.dev0+61ef5cf"
```

`how_to_release.rst` tells the maintainer to add `.dev0` to the version. Thus this defect occurs at the next release. Also, `pyedflib.__version__` becomes different from the version in the wheel metadata. The old `setup.py` used `version=get_version_info()[0]`, and the two values agreed.

**3. The wheel contains the build sources.**

In `pyproject.toml`, `include-package-data` is true by default, and `MANIFEST.in` has `recursive-include pyedflib *`. Thus the wheel now contains `_pyedflib.pyx`, `edf.pxi`, `c_edf.pxd`, and all files in `_extensions/c/`. This adds approximately 40 kB. Set `include-package-data = false`. Then `[tool.setuptools.package-data]` becomes effective. At this time that section has no effect.

### Discussion: keep a small `setup.py`

Issue #225 asks for this method: "Move to `pyproject.toml` and make `setup.py` a thin wrapper." This PR deletes `setup.py` instead.

Put all the metadata in `pyproject.toml` with PEP 621, as this PR does. Keep `build-backend = "setuptools.build_meta"`. Then add a small `setup.py` that contains only `ext_modules` and `libraries`. Setuptools supports this configuration. Setuptools reads the `[project]` table, and those values have priority. Thus the `setup()` call must not repeat them.

This method deletes approximately 110 of the 386 lines in `_custom_build.py`. (My first message said 180 lines. 110 is more accurate.) These parts go away:

- the eight PEP 517 hooks (`build_wheel`, `build_sdist`, `build_editable`, `get_requires_for_build_*`, `prepare_metadata_*`) — approximately 40 lines
- `_orig_finalize`, `_patched_finalize`, and the `_pyedflib_injected` flag — approximately 20 lines
- `_LazyNumpyExtension` — approximately 20 lines
- `_prepare_build()`, the `_stbm` import, and the backend docstring
- `backend-path = ["."]` in `pyproject.toml`

The other parts move into `setup.py` with almost no change: the version reader, `_git_revision()`, `get_version_info()`, `write_version_py()`, the Windows patch, the numpy include function, and the `c_edf` and `Extension` definitions.

Three details make the small `setup.py` less complex than the backend:

1. `write_version_py()` becomes one call at module level, before `setup()`. At this time the code calls it from five different hooks.
2. `_LazyNumpyExtension` is not necessary. numpy is already in `[build-system] requires`, thus pip installs numpy before the build starts. Also, `cythonize()` reads `include_dirs` immediately, thus the delay gives no benefit.
3. `setup.py` defines `ext_modules` and `libraries` as arguments to `setup()`. Thus you do not patch `Distribution.finalize_options`.

The patch of `finalize_options` has these effects:

- `finalize_options` runs for each command. Thus `cythonize()` runs more than one time in each build.
- The code calls `get_libraries()` two times. Thus `_patch_edflib_utf8()` runs two times on Windows.

Keep the removal of `develop_build_clib`. That change is correct, and it is independent of the backend question. `build_ext` builds `c_edf` because `has_c_libraries()` is true. I confirmed this with `pip install -e .`. Thus a small `setup.py` is not the old file. It has no develop hack, no metadata block, and no lazy numpy class.

DimitriPapadopoulos:
Perhaps keeping a `setup.py` file is indeed easier. Let me try.

DimitriPapadopoulos:
See #324 for an alternative where we keep a minimal `setup.py`. It's indeed simpler, I suggest we switch to that alternative solution.

skjerns:
completed in #324

PR #299, opened 2026-07-19
Start applying ruff/Pylint rules (`PLR`)

DimitriPapadopoulos:


skjerns:
Wow, I first misread the diff and thought the current state was what you would propose and thought "why would ruff enforce such ugly code"? But then I saw it was the current state of the repo. good change!

PR #300, opened 2026-07-20
Add pad_with option to control padding of the final data record

skjerns:
Follow-up to #286, as discussed there.

`EdfWriter(..., pad_with=...)` controls the value used to fill up an incomplete last data record: a number (default `0`, unchanged behaviour) or `'last'` to repeat each signal's last sample. Zero-padding drags the final partial record to the zero line, which shows up as a step discontinuity / spectral edge artifact at the end of biopotential recordings; `'last'` avoids that.

```python
w = pyedflib.EdfWriter('out.edf', 1, buffered=True, pad_with='last')
```

The default stays `0` for backward compatibility — `'last'` is opt-in.

Details worth reviewing:
- `pad_with` applies to both the buffered flush on `close()` and the regular `writeSamples()` tail block, so it behaves the same either way.
- With `pad_with='last'`, the last sample of each channel is tracked separately rather than read from the leftover buffer. With differing sample frequencies a channel can end exactly on a record boundary while another still has leftovers, and reading from the (then empty) buffer would pad a whole record with zeros.
- `digital=True` requires an integer `pad_with` (raises `TypeError`) instead of silently truncating e.g. `2.7` to `2`. A pad value outside the channel range warns that edflib will clip it.
- numpy scalars (`pad_with=signal[-1]`) are accepted; `bool` is rejected.
- Validation happens before the file is opened, so a rejected argument no longer leaves an empty file behind. `close()` now tolerates a partially constructed object, which also fixes a pre-existing `AttributeError` in `__del__` when `__init__` raised (e.g. on the `open_file_writeonly` error path).

6 new tests, one per case above. Full suite passes (82).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyedflib/edfwriter.py
+++ b/pyedflib/edfwriter.py
@@ -286,7 +286,8 @@ class EdfWriter:
         self.close()
 
     def __init__(self, file_name: str, n_channels: int, file_type: int = FILETYPE_EDFPLUS,
-                 buffered: bool = False) -> None:
+                 buffered: bool = False,
+                 pad_with: Union[int, float, str] = 0) -> None:
         """Initialises an EDF file at file_name.
         file_type is one of
             edflib.FILETYPE_EDFPLUS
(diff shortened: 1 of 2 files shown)


skjerns:
@aagisto-maker can you check this works for your use-case?

aagisto-maker:
Confirmed on the merged master (with the enumerate / `_get_pad_value` refactors in) — works for my use-case. Buffered EMG at 1 kHz, 3 channels, 100-sample blocks with a partial final record:

- Default (`pad_with=0`): final partial record zero-padded — backward-  compatible.
- `pad_with='last'`: the trailing pad takes each channel's last acquired value; with a signal ending well off zero, the ~1.5 mV zero-pad step disappears (step ≈ 0). Real samples untouched, the out-of-range warning fires, and the buffered/pad tests pass here.

I'll use `pad_with='last'` for the EMG streams. Thanks for getting this in so fast.

PR #303, opened 2026-07-20
Apply ruff rule RUF012

DimitriPapadopoulos:


skjerns:
I'm not sure the `utils` stuff is actually in use anywhere. As far as I can see all of the scripts are deprecated, based on Python 2, calling non-used toolchains etc. We can just get rid of it, except `authors.py`, but probably even that, need to check.

skjerns:
I've removed the `utils` in #306 , so this is deprecated now, except for the `RUF012`. Don't think applies to our code much, but no harm in keeping it

PR #315, opened 2026-07-28
Fix reading and writing subsecond part of start time

augeles:
Fixes the incorrectly read and written subsecond part in start timeand corrects the test case.

As noted in the [EDFlib documentation](https://www.teuniz.net/edflib/doc/edflib_8h.html#a5dc9e8730281c4937745902b70d86dd7), the subsecond start time is expressed in units of 100 nanoseconds, so the conversion factor to microseconds expected by _datetime_ should be 10 not 100.
Also the corresponding pytest checked against _2020-01-24 04:05:56.039453_ instead of the correct _2020-01-24 04:05:56.394531_.

### DIFF
--- a/pyedflib/edfreader.py
+++ b/pyedflib/edfreader.py
@@ -434,8 +434,8 @@ class EdfReader(CyEdfReader):
         >>> f.close()
 
         """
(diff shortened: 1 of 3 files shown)


skjerns:
oh, indeed! thanks for catching that!

PR #324, opened 2026-08-03
Migrate to pyproject.toml with PEP 621 metadata

### LINKED ISSUE #225: setup.py → pyproject.toml

DimitriPapadopoulos:
Also apply PEP 639 and 735 to update metadata.

Alternative to #298: it seems simpler to keep a minimal `setup.py`. Also, let's defer version handling with `setuptools-scm` to a later PR.

Fixes https://github.com/holgern/pyedflib/issues/225.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -21,10 +21,8 @@ jobs:
         python-version: 3.9
     - name: Generate coverage report
       run: |
-        pip install pytest
-        pip install pytest-cov
         pip install -e . --group test
-        pytest --cov=./pyedflib/ --cov-report=xml --cov-config=.coveragerc
+        python -m pytest --cov=./pyedflib/ --cov-report=xml --cov-config=.coveragerc
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v7
       with:
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -23,8 +23,7 @@ jobs:
       run: |
         pip install pytest
         pip install pytest-cov
-        pip install -e .
-        pip install -r requirements-test.txt
+        pip install -e . --group test
         pytest --cov=./pyedflib/ --cov-report=xml --cov-config=.coveragerc
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v7
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,5 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install packages
       run: |
-        pip install pytest
         pip install -e . --group test
-        pytest
+        python -m pytest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,5 @@ jobs:
     - name: Install packages
       run: |
         pip install pytest
-        pip install -e .
-        pip install -r requirements-test.txt
+        pip install -e . --group test
         pytest
--- a/.github/workflows/testz_numpy_2_compatibility.yml
+++ b/.github/workflows/testz_numpy_2_compatibility.yml
@@ -20,6 +20,5 @@ jobs:
       run: |
         pip install numpy==1.24
         pip install pytest
-        pip install -e .
-        pip install -r requirements-test.txt
+        pip install -e . --group test
         pytest
--- a/.github/workflows/testz_numpy_2_compatibility.yml
+++ b/.github/workflows/testz_numpy_2_compatibility.yml
@@ -19,6 +19,5 @@ jobs:
     - name: Install packages
       run: |
         pip install numpy==1.24
-        pip install pytest
         pip install -e . --group test
-        pytest
+        python -m pytest
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,13 +6,11 @@ build:
   os: ubuntu-26.04
   tools:
     python: "3.14"
+  jobs:
+    install:
+      - pip install --upgrade pip
+      # Install pyedflib itself so autodoc can import the compiled extension.
+      - pip install --group docs .
 
 sphinx:
   configuration: doc/source/conf.py
-
-python:
-  install:
-    # Since the install step is overridden, pip is no longer updated automatically.
-    - pip install --upgrade pip
-    - pip install --group docs .
-    # Install pyedflib itself so autodoc can import the compiled extension.
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,17 +3,16 @@
 version: 2
 
 build:
-  os: ubuntu-24.04
+  os: ubuntu-26.04
   tools:
-    python: "3.12"
+    python: "3.14"
 
 sphinx:
   configuration: doc/source/conf.py
 
 python:
   install:
-    # Documentation build dependencies (sphinx, numpydoc, ...).
-    - requirements: requirements-doc.txt
+    # Since the install step is overridden, pip is no longer updated automatically.
+    - pip install --upgrade pip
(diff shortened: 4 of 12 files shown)


pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
-    "cython"
+    "cython>=0.29",
 ]
 build-backend = "setuptools.build_meta"
 
+
+[project]
+name = "pyEDFlib"
+maintainers = [
+    { name = "Holger Nahrstaedt", email = "nahrstaedt@gmail.com" },
```
skjerns, line 13: Feel free to add us both here! I don't think it's worth another PR

pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
+]
+authors = [
+    { name = "Holger Nahrstaedt", email = "nahrstaedt@gmail.com" },
+]
+description = "library to read/write EDF+/BDF+ files"
+readme = "README.rst"
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
+requires-python = ">=3.9"
```
skjerns, line 24:
```suggestion
requires-python = ">=3.11"
```

we raised it to that in a previous PR, if I remember?

pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
+    "Intended Audience :: Developers",
+    "Intended Audience :: Education",
+    "Intended Audience :: Science/Research",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: C",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
skjerns, line 36:
```suggestion
```

pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
+test = [
+    "coverage",
+    "dateparser",
+    "pytest",
+    "ruff",
+    "tqdm",
+]
+dev = [
+    "tox",
+    "typing_extensions; python_version < '3.11'",
```
skjerns:
```suggestion
```

when we generally raise to 3.11, we don't need that anymore

.github/workflows/tests.yml
```diff
@@ -20,6 +20,5 @@ jobs:
     - name: Install packages
       run: |
         pip install pytest
```
skjerns:
```suggestion
```

is already installed via the line below!

.github/workflows/codecov.yml
```diff
@@ -23,8 +23,7 @@ jobs:
       run: |
         pip install pytest
         pip install pytest-cov
```
skjerns: is already installed via the `--group test`

pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
+dynamic = ["version"]
+dependencies = [
+    "numpy>=1.9.1",
+]
+
+[dependency-groups]
+test = [
+    "coverage",
+    "dateparser",
+    "pytest",
```
skjerns, line 60:
```suggestion
    "pytest",
    "pytest-cov",
```

in hope that at some point the coverage will work again :-/

.github/workflows/testz_numpy_2_compatibility.yml
```diff
@@ -20,6 +20,5 @@ jobs:
       run: |
         pip install numpy==1.24
         pip install pytest
```
skjerns:
```suggestion
```

pyproject.toml
DimitriPapadopoulos: We actually raised this value from 3.8 to 3.9 in #293. Python 3.10 hasn't reached EOL yet.

.readthedocs.yaml
```diff
@@ -3,17 +3,16 @@
 version: 2
 
 build:
-  os: ubuntu-24.04
+  os: ubuntu-26.04
   tools:
-    python: "3.12"
+    python: "3.14"
```
skjerns, line 8:
```suggestion
    python: "3.14"
  jobs:
    install:
      - pip install --upgrade pip
      # Install pyedflib itself so autodoc can import the compiled extension.
      - pip install --group docs .
```

Claude flagged this file as invalid

> python.install entries must be mappings (requirements: <path>, or method:/path:). Plain shell strings are not accepted — confirmed against RTD's config reference

therefore, the commands must be under the `jobs` path apparently. Please verify the yaml is valid after my suggestion, it's quite finicky with the indents

.readthedocs.yaml
```diff
@@ -3,17 +3,16 @@
...
 sphinx:
   configuration: doc/source/conf.py
 
 python:
   install:
-    # Documentation build dependencies (sphinx, numpydoc, ...).
-    - requirements: requirements-doc.txt
+    # Since the install step is overridden, pip is no longer updated automatically.
+    - pip install --upgrade pip
+    - pip install --group docs .
```
skjerns: this must be removed then (see above)

pyproject.toml
```diff
@@ -1,13 +1,97 @@
...
+    "ruff",
+    "tqdm",
+]
+dev = [
+    "tox",
+    "typing_extensions; python_version < '3.11'",
+]
```
skjerns:
```suggestion
```

I think we can remove the dev-dependencies and only rely on the `test` group.
 `tox` is no longer used anyway

skjerns:
Great work! I think this is the much better solution.

I've made some comments, let me know what you think. Also, we can probably remove the `dev` group for the requirements and rely solely on `test`, as the `tox` dependency is no longer there. 

or alternatively, make the dev group an inclusion of `test` + `docs`

``` python
dev = [
{ include-group = "test" },
{ include-group = "docs" },
]
```

pyproject.toml
skjerns: Aaah true!

pyproject.toml
skjerns: In this case, we need to include the `"typing_extensions; python_version < '3.11'",` in the general dependency group, as otherwise it will raise an error on Python `3.9` when not installing via `dev/test`

DimitriPapadopoulos:
As suggested, I have made the `dev` group a combination of `test` and `docs` instead of removing it, as it appears to be an expected default:
https://learn.scientific-python.org/development/guides/packaging-compiled/#development-dependencies

.readthedocs.yaml
DimitriPapadopoulos: I'm not sure how to test it works. We lack a test for documentation. Let me have a look at it again.

DimitriPapadopoulos:
I've applied all changes, except the Python 3.9 → 3.11 one.

pyproject.toml
```diff
@@ -1,13 +1,100 @@
...
+    "EDFlib",
+    "European data format",
+    "EDF",
+    "BDF",
+    "EDF+",
+    "BDF+",
+]
+dynamic = ["version"]
+dependencies = [
+    "numpy>=1.9.1",
```
skjerns, line 53:
```suggestion
    "numpy>=1.9.1",
    "typing_extensions; python_version < '3.11'",
```

we need to add it as a general dependency, as 3.9 users otherwise get an import error. it can then be removed from the `test` group

skjerns:
Thanks! One more small comment and then we're good to go :)

let me know what you think about the `yaml`

pyproject.toml
DimitriPapadopoulos: Indeed, it's a runtime dependency.

skjerns:
merged

great work, thanks!

PR #325, opened 2026-08-04
Run latest MyPy on code base

DimitriPapadopoulos:
It appears MyPy wasn't running any more in tests and CI!

How about we use pre-commit for MyPy and Ruff?


skjerns:
precommit sounds good!

will those be activated automatically, or do the devs have to activate them somehow? If yes (and not possible to do automatically), we'll also need to update the contribution instructions

DimitriPapadopoulos:
After adding file `.pre-commit-config.yaml`, [pre-commit.ci](https://pre-commit.ci/) needs to be enabled in project settings. An alternative would be [pre-commit.ci lite](https://pre-commit.ci/lite.html) which runs as a GitHub action.

DimitriPapadopoulos:
On developer machines, pre-commit is not enabled automatically. We need to add instructions:
https://pre-commit.com/#install

DimitriPapadopoulos:
Need to merge #328 first to enable pre-commit.

skjerns:
is it ready to merge now?

DimitriPapadopoulos:
Ready to merge.

